PR #27, opened 2022-11-15
Add User endpoints

samleeflang:
Add User endpoints (CRUD) with the JSON:API specification


src/main/java/eu/dissco/backend/controller/UserController.java
```diff
@@ -0,0 +1,124 @@
...
+@RestController
+@RequestMapping("/api/v1/users")
+@RequiredArgsConstructor
+public class UserController {
+
+  private static final String SELF_LINK = "https://sandbox.dissco.tech/api/v1/users/";
+
+  private final UserService service;
+
+  private static void checkAuthorisation(String tokenId, String id) throws ForbiddenException {
```
southeo, line 42: Just to clarify: @PreAuthorize("isAuthenticated") checks if the user has permission to access the API at all, and the checkAuthorisation function checks the user's specific permissions for each action?

src/main/java/eu/dissco/backend/controller/UserController.java
```diff
@@ -0,0 +1,124 @@
...
+  @PreAuthorize("isAuthenticated()")
+  @ResponseStatus(HttpStatus.CREATED)
+  @PostMapping
+  public ResponseEntity<JsonApiWrapper> createNewUser(Authentication authentication,
+      @RequestBody JsonApiWrapper request)
+      throws JsonProcessingException, ConflictException, ForbiddenException {
+    var tokenId = getNameFromToken(authentication);
+    log.info("User: {} has requested to update user information of: {}",
+        tokenId, request.data().id());
+    checkAuthorisation(tokenId, request.data().id());
```
southeo, line 57: Can you explain how comparing the user tokenId to request.data().id ensures the user is authorized? Could the request id not be manipulated?

src/main/java/eu/dissco/backend/controller/UserController.java
samleeflang: Yes there is a hasRole as well to check if a user has a certain role and I believe a couple of others to check group.

src/main/java/eu/dissco/backend/controller/UserController.java
samleeflang: A different user could use the userId of another user in the request and then delete that user :smiling_face_with_tear: 
So let's say I have your userId (which are public) and send a request to the server with a delete for your userId. That would be totally acceptable if we didn't do this check as you are a authorized user. So that's where the check comes in.
- Create/Update/Delete a user can only do for itself (although delete is a bit strange)
- Find all authorized users can do (might become public later, not sure about that)

samleeflang: Example:
DELETE -> `localhost:8080/api/v1/users/e2befba6-9324-4bb4-9f41-d7dfae4a44b0`
For my user with the bearer token for user `adf294ba-bb03-4962-8042-a37f1648458e`

src/test/java/eu/dissco/backend/TestUtils.java
```diff
@@ -0,0 +1,54 @@
...
+    return new JsonApiData(id, TYPE, MAPPER.valueToTree(givenUser()));
+  }
+
+  public static User givenUser() {
+    return new User(
+        "Test",
+        "User",
+        "test@gmail.com",
+        "https://orcid.org/0000-0002-XXXX-XXXX",
+        "https://ror.org/0566bfb96"
```
TomDijkema, line 51: Does the ROR foundation maybe have a default test ROR identifier? Instead of using the Naturalis one. Probably not very important however...

src/test/resources/db/migration/V1__init_db.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE public.new_user (
+	id text NOT NULL,
+	first_name text NULL,
+	last_name text NULL,
```
TomDijkema, line 4: Keycloak calls these field in plural like: 'first_names' and 'last_names', how do you think we are going to handle multiple first or last names?

src/test/java/eu/dissco/backend/TestUtils.java
samleeflang: No this identifier isn't called anywhere so don't see an issue there. Could also use "XXX" but this way it at least looked like an ROR

src/test/resources/db/migration/V1__init_db.sql
samleeflang: There can be a space in the text field so "Tom Jan Dijkema Van Dijk" Would split into: "Tom Jan" and "Dijkema Van Dijk". Don't really see an issue there, same as Elvis is doing it.

src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -21,6 +23,10 @@
 @RequiredArgsConstructor
 public class SpecimenService {
 
+  private final Map<String, String> prefixMap = Map.of("dct", "http://purl.org/dc/terms/", "dwc",
+      "http://rs.tdwg.org/dwc/terms/", "abcd", "https://abcd.tdwg.org/terms/", "abcd-efg",
+      "https://terms.tdwg.org/wiki/ABCD_EFG/");
+
```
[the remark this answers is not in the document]
samleeflang: Not sure these are I believe the ontologies that we support but you might be right.

src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -96,19 +102,27 @@ private JsonNode generatePrimaryData(DigitalSpecimen digitalSpecimen) {
     return primarySpecimenData;
   }
 
-  private JsonNode generateContext() {
+  private JsonNode generateContext(JsonNode data) {
+    var prefixes = determinePrefixes(data);
     var node = mapper.createObjectNode();
     node.put("ods", "http://github.com/DiSSCo/openDS/ods-ontology/terms/");
```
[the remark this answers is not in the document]
samleeflang: Yes these are always added I think that is way I didn't add them. Might add them as well, don't see an iddue

src/main/java/eu/dissco/backend/database/jooq/tables/records/NewUserRecord.java
```diff
@@ -0,0 +1,368 @@
...
+import org.jooq.Record8;
+import org.jooq.Row8;
+import org.jooq.impl.UpdatableRecordImpl;
+
+
+/**
+ * This class is generated by jOOQ.
+ */
+@SuppressWarnings({ "all", "unchecked", "rawtypes" })
+public class NewUserRecord extends UpdatableRecordImpl<NewUserRecord> implements Record8<String, String, String, String, String, String, Instant, Instant> {
```
southeo, line 22: Is this not recursive? NewUserRecord extends Updatable...<**NewUserRecord**>?

src/main/java/eu/dissco/backend/database/jooq/tables/NewUser.java
```diff
@@ -0,0 +1,173 @@
...
+import org.jooq.impl.DSL;
+import org.jooq.impl.SQLDataType;
+import org.jooq.impl.TableImpl;
+
+
+/**
+ * This class is generated by jOOQ.
+ */
+@SuppressWarnings({ "all", "unchecked", "rawtypes" })
+public class NewUser extends TableImpl<NewUserRecord> {
```
southeo, line 34: This class extends NewUserRecord, but NewUserRecord imports NewUser. Is that not circular?
I'm not clear on the relationship between these two classes

src/main/java/eu/dissco/backend/service/SpecimenService.java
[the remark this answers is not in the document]
samleeflang: Will keep this here for now as externalizing it would be a bit of effort. Let's make this changes if we add support for other ontologies.

src/main/java/eu/dissco/backend/repository/UserRepository.java
```diff
@@ -0,0 +1,77 @@
...
+@Repository
+@RequiredArgsConstructor
+public class UserRepository {
+
+  private final DSLContext context;
+  private final Map<String, TableField<NewUserRecord, String>> attributeMapping = Map.of(
+      "firstName", NEW_USER.FIRST_NAME,
+      "lastName", NEW_USER.LAST_NAME,
+      "email", NEW_USER.EMAIL,
+      "orchid", NEW_USER.ORCHID,
```
southeo, line 29: "Orchid" = the plant
"Orcid" = open researcher and contributer id
You probably want to change this

src/main/java/eu/dissco/backend/database/jooq/tables/records/NewUserRecord.java
samleeflang: This code is generated by JOOQ

src/main/java/eu/dissco/backend/database/jooq/tables/NewUser.java
samleeflang: This code is generated by JOOQ

src/main/java/eu/dissco/backend/repository/UserRepository.java
samleeflang: :+1:

samleeflang: Fixed

src/main/java/eu/dissco/backend/service/SpecimenService.java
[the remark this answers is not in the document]
samleeflang: Fixed, added them to the map

src/test/java/eu/dissco/backend/repository/UserRepositoryIT.java
```diff
@@ -0,0 +1,86 @@
...
+import static eu.dissco.backend.TestUtils.givenUser;
+import static eu.dissco.backend.database.jooq.Tables.NEW_USER;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class UserRepositoryIT extends BaseRepositoryIT {
```
southeo, line 14: Here's how I understand this: 

1) BaseRepositoryIT sets up a container to run the repository in. The BaseRepositoryIT container is based off of a library (PostgreSQLContainer.IMAGE). 
2) In this class, you can make test calls to the repository and have it behave normally, creating/updating/deleting records without impacting the database.

samleeflang: Yes so the baseRepository kicks of a couple things:
- It creates a Postgres container
- It creates a datasource which can be used to interact with the container
- It checks if the database/source works with a test query
- It populates to database with the sql in the db.migrations, for this we use the Flyway library
- It create the JOOQ context object which we use in the repository

Then in this class when we start it all these things will be triggered so that when we want to test anything we will have a database with a context. We can create a repository object with an actual contact (not a mock, so it is an integration test) and start running actual methods against an actual database.

southeo: Noted, thanks!

southeo (approved):
Nice, thanks for clarifying on points.

TomDijkema (approved):

PR #31, opened 2023-02-23
Feature/use mongo versions

samleeflang:
Versions have been removed from database.
Database now only holds latest version making easy joins possible.
This is now reflected in the code.

Mongodb will provide the version information.

### DIFF
--- a/src/main/java/eu/dissco/backend/configuration/ApplicationConfiguration.java
+++ b/src/main/java/eu/dissco/backend/configuration/ApplicationConfiguration.java
@@ -2,6 +2,7 @@ package eu.dissco.backend.configuration;
 
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.text.SimpleDateFormat;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
@@ -12,6 +13,7 @@ public class ApplicationConfiguration {
   @Bean
   public ObjectMapper objectMapper() {
     var mapper = new ObjectMapper().findAndRegisterModules();
+    mapper.setDateFormat(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS"));
     mapper.setSerializationInclusion(Include.NON_NULL);
     return mapper;
   }
--- a/src/main/java/eu/dissco/backend/controller/AnnotationController.java
+++ b/src/main/java/eu/dissco/backend/controller/AnnotationController.java
@@ -1,9 +1,11 @@
 package eu.dissco.backend.controller;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import eu.dissco.backend.domain.AnnotationRequest;
 import eu.dissco.backend.domain.AnnotationResponse;
 import eu.dissco.backend.domain.JsonApiListResponseWrapper;
 import eu.dissco.backend.exceptions.NoAnnotationFoundException;
+import eu.dissco.backend.exceptions.NotFoundException;
 import eu.dissco.backend.service.AnnotationService;
 import java.io.IOException;
 import java.util.List;
@@ -76,7 +78,8 @@ public class AnnotationController {
 
   @GetMapping(value = "/{prefix}/{postfix}/{version}", produces = MediaType.APPLICATION_JSON_VALUE)
   public ResponseEntity<AnnotationResponse> getAnnotation(@PathVariable("prefix") String prefix,
-      @PathVariable("postfix") String postfix, @PathVariable("version") int version) {
+      @PathVariable("postfix") String postfix, @PathVariable("version") int version)
+      throws JsonProcessingException, NotFoundException {
     var id = prefix + '/' + postfix;
     log.info("Received get request for annotation: {} with version: {}", id, version);
     var annotation = service.getAnnotationVersion(id, version);
@@ -165,7 +168,7 @@ public class AnnotationController {
   @ResponseStatus(HttpStatus.OK)
   @GetMapping(value = "/{prefix}/{postfix}/versions", produces = MediaType.APPLICATION_JSON_VALUE)
   public ResponseEntity<List<Integer>> getAnnotationByVersion(@PathVariable("prefix") String prefix,
-      @PathVariable("postfix") String postfix) {
+      @PathVariable("postfix") String postfix) throws NotFoundException {
     var id = prefix + '/' + postfix;
     log.info("Received get request for versions of annotation with id: {}", id);
     var versions = service.getAnnotationVersions(id);
--- a/src/main/java/eu/dissco/backend/controller/AnnotationController.java
+++ b/src/main/java/eu/dissco/backend/controller/AnnotationController.java
@@ -77,12 +77,12 @@ public class AnnotationController {
   }
 
   @GetMapping(value = "/{prefix}/{postfix}/{version}", produces = MediaType.APPLICATION_JSON_VALUE)
-  public ResponseEntity<AnnotationResponse> getAnnotation(@PathVariable("prefix") String prefix,
+  public ResponseEntity<AnnotationResponse> getAnnotationByVersion(@PathVariable("prefix") String prefix,
       @PathVariable("postfix") String postfix, @PathVariable("version") int version)
       throws JsonProcessingException, NotFoundException {
     var id = prefix + '/' + postfix;
     log.info("Received get request for annotation: {} with version: {}", id, version);
-    var annotation = service.getAnnotationVersion(id, version);
+    var annotation = service.getAnnotationByVersion(id, version);
     return ResponseEntity.ok(annotation);
   }
 
@@ -167,7 +167,7 @@ public class AnnotationController {
 
   @ResponseStatus(HttpStatus.OK)
   @GetMapping(value = "/{prefix}/{postfix}/versions", produces = MediaType.APPLICATION_JSON_VALUE)
-  public ResponseEntity<List<Integer>> getAnnotationByVersion(@PathVariable("prefix") String prefix,
+  public ResponseEntity<List<Integer>> getAnnotationVersions(@PathVariable("prefix") String prefix,
       @PathVariable("postfix") String postfix) throws NotFoundException {
     var id = prefix + '/' + postfix;
     log.info("Received get request for versions of annotation with id: {}", id);
--- a/src/main/java/eu/dissco/backend/controller/DigitalMediaObjectController.java
+++ b/src/main/java/eu/dissco/backend/controller/DigitalMediaObjectController.java
@@ -1,8 +1,11 @@
 package eu.dissco.backend.controller;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import eu.dissco.backend.domain.AnnotationResponse;
 import eu.dissco.backend.domain.DigitalMediaObject;
 import eu.dissco.backend.domain.JsonApiListResponseWrapper;
 import eu.dissco.backend.domain.JsonApiWrapper;
+import eu.dissco.backend.exceptions.NotFoundException;
 import eu.dissco.backend.service.DigitalMediaObjectService;
 import java.util.List;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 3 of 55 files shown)


src/main/java/eu/dissco/backend/configuration/ApplicationConfiguration.java
```diff
@@ -12,6 +13,7 @@ public class ApplicationConfiguration {
   @Bean
   public ObjectMapper objectMapper() {
     var mapper = new ObjectMapper().findAndRegisterModules();
+    mapper.setDateFormat(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS"));
     mapper.setSerializationInclusion(Include.NON_NULL);
```
southeo, line 17: In the handle server, I have this as 'yyyy-MM-dd'T'HH:mm:ss.SSS'Z'' (with the Z at the end) to indicate UTC time. That's the pattern from our discussion earlier.

src/main/java/eu/dissco/backend/controller/AnnotationController.java
```diff
@@ -76,7 +78,8 @@ public ResponseEntity<JsonApiListResponseWrapper> getLatestAnnotationsJsonRespon
 
   @GetMapping(value = "/{prefix}/{postfix}/{version}", produces = MediaType.APPLICATION_JSON_VALUE)
   public ResponseEntity<AnnotationResponse> getAnnotation(@PathVariable("prefix") String prefix,
-      @PathVariable("postfix") String postfix, @PathVariable("version") int version) {
+      @PathVariable("postfix") String postfix, @PathVariable("version") int version)
+      throws JsonProcessingException, NotFoundException {
```
southeo, line 82: We should really change "postfix" to "suffix" at some point...

src/main/java/eu/dissco/backend/controller/DigitalMediaObjectController.java
```diff
@@ -76,22 +79,19 @@ public ResponseEntity<JsonApiWrapper> getMultiMediaByIdJsonResponse(
...
-      @RequestParam(defaultValue = DEFAULT_PAGE_SIZE) int pageSize,
-      HttpServletRequest request) {
+  public ResponseEntity<List<AnnotationResponse>> getAnnotationsById(
+      @PathVariable("prefix") String prefix, @PathVariable("postfix") String postfix) {
     var id = prefix + '/' + postfix;
-    String path = SANDBOX_URI + request.getRequestURI();
     log.info("Received get request for annotations on digitalMedia with id: {}", id);
-    var annotations = service.getAnnotationsOnDigitalMediaObject(id, path, pageNumber, pageSize);
+    var annotations = service.getAnnotationsOnDigitalMediaObject(id);
     return ResponseEntity.ok(annotations);
```
southeo, line 87: Shouldn't this be in JsonApi?

src/main/java/eu/dissco/backend/controller/UserController.java
```diff
@@ -97,21 +97,6 @@ public ResponseEntity<Void> deleteUser(Authentication authentication,
     return ResponseEntity.status(HttpStatus.NO_CONTENT).build();
   }
 
-  @ExceptionHandler(ConflictException.class)
```
southeo, line 100: Why get rid of these?

src/main/java/eu/dissco/backend/repository/DigitalMediaObjectRepository.java
```diff
@@ -29,118 +27,48 @@ public class DigitalMediaObjectRepository {
...
-        .from(NEW_ANNOTATION)
-        .where(NEW_ANNOTATION.TARGET_ID.eq(mediaId))
-        .offset(offset)
-        .limit(pageSize)
-        .fetch(this::mapAnnotationToJsonApiData);
-  }
-
-
   public List<DigitalMediaObject> getDigitalMediaForSpecimen(String id) {
     return context.select(NEW_DIGITAL_MEDIA_OBJECT.asterisk())
```
southeo, line 46: I think this function and the previous functions' names should be swapped. This function gets latest digital media objects, the previous function gets media for a given specimen.

src/main/java/eu/dissco/backend/repository/MongoRepository.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.stereotype.Repository;
+
+@Repository
+@RequiredArgsConstructor
+public class MongoRepository {
+
+  private final MongoDatabase database;
+  private final ObjectMapper mapper;
+
+  public JsonNode getVersion(String id, int version, String collectionName)
```
southeo, line 23: Small thing: consider calling this getByVersion? You're not retrieving the version, you're using the version to retrieve a record. 
I don't see any uses for this yet

src/main/java/eu/dissco/backend/repository/SpecimenRepository.java
```diff
@@ -22,23 +22,18 @@ public class SpecimenRepository {
...
     return context.select(NEW_DIGITAL_SPECIMEN.asterisk())
         .from(NEW_DIGITAL_SPECIMEN)
         .offset(offset)
         .limit(pageSize)
         .fetch(this::mapper);
   }
 
-  public DigitalSpecimen getLatestSpecimenById(String id) {
+  public DigitalSpecimen getSpecimenById(String id) {
     return context.select(NEW_DIGITAL_SPECIMEN.asterisk())
```
southeo, line 34: You're still getting the latest specimen by Id, I'd keep the original name (is consistent with annotation repository too)

src/main/java/eu/dissco/backend/service/AnnotationService.java
```diff
@@ -123,17 +127,36 @@ private AnnotationResponse mapResponseToAnnotationResponse(JsonNode response) {
         Instant.ofEpochSecond(annotation.get("generated").asLong()), null);
   }
 
-  public AnnotationResponse getAnnotationVersion(String id, int version) {
-    return repository.getAnnotationVersion(id, version);
+  public AnnotationResponse getAnnotationVersion(String id, int version)
+      throws JsonProcessingException, NotFoundException {
```
southeo, line 131: getAnnotationByVersion ?

src/test/java/eu/dissco/backend/repository/ElasticSearchRepositoryIT.java
```diff
@@ -54,7 +61,7 @@ class ElasticSearchRepositoryIT {
   private static final String ELASTICSEARCH_PASSWORD = "s3cret";
   private static final String CREATED_ALT = "2022-09-02T09:59:24Z";
   private static final ElasticsearchContainer container = new ElasticsearchContainer(
-      ELASTIC_IMAGE).withExposedPorts(9200).withPassword(ELASTICSEARCH_PASSWORD);
+      ELASTIC_IMAGE).withPassword(ELASTICSEARCH_PASSWORD);
```
southeo, line 64: Why remove withExposedPorts()?

src/test/java/eu/dissco/backend/repository/ElasticSearchRepositoryIT.java
```diff
@@ -187,7 +200,7 @@ void testGetLatestSpecimenSecondPage() throws IOException {
 
     // When
     var responseReceived = repository.getLatestSpecimen(pageNumber, pageSize);
-    var allElems = repository.getLatestAnnotations(1, pageSize * 2);
+    repository.getLatestAnnotations(1, pageSize * 2);
```
southeo, line 204: Should delete this line altogether

src/test/java/eu/dissco/backend/repository/SpecimenRepositoryIT.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Test
+  void testGetSpecimenById() throws JsonProcessingException {
+    // Given
+    populateSpecimenTable();
+
+    // When
+    var result = repository.getSpecimenById("20.5000.1025/ABC-123-XY3");
+
```
southeo, line 51: Should use an id from testUtils

src/main/java/eu/dissco/backend/controller/DigitalMediaObjectController.java
```diff
@@ -101,21 +101,11 @@ public ResponseEntity<List<Integer>> getDigitalMediaVersions(
...
       @PathVariable("prefix") String prefix, @PathVariable("postfix") String postfix,
-      @PathVariable("version") int version) {
+      @PathVariable("version") int version) throws JsonProcessingException, NotFoundException {
     var id = prefix + '/' + postfix;
     log.info("Received get request for digital media: {} with version: {}", id, version);
     var digitalMedia = service.getDigitalMediaVersion(id, version);
     return ResponseEntity.ok(digitalMedia);
   }
 
-  @GetMapping(value = "/{prefix}/{postfix}/{version}/json", produces = MediaType.APPLICATION_JSON_VALUE)
```
southeo, line 111: Why delete this?

southeo (commented):
Looks good, those database calls look so much nicer without those joins. Just a few clarifying questions.

src/main/java/eu/dissco/backend/configuration/ApplicationConfiguration.java
samleeflang: Ah I don't think that is correct. You hardcode the Z then at the end which doesn't always have to be true. I think then it should be `yyyy-MM-dd'T'HH:mm:ss.SSSXXX`. However it still gives me five behind the second so for some reason it is not working as expected and I have to dig into this some more to get the formatting right

src/main/java/eu/dissco/backend/controller/AnnotationController.java
samleeflang: Yes item is already on the backlog, see 
https://naturalis.atlassian.net/browse/DD-215

src/main/java/eu/dissco/backend/controller/DigitalMediaObjectController.java
samleeflang: I reverted this because it changed from the methods in the MediaObjectService and Repo to the annotation. It now makes use of a function which wasn't converted to JSON:API. Changing this will be part of:
https://naturalis.atlassian.net/browse/DD-240

src/main/java/eu/dissco/backend/controller/UserController.java
samleeflang: I didn't get rid of these but I made them generic for all classes by putting them in the `RestResponseEntityExceptionHandler.class`

src/main/java/eu/dissco/backend/repository/MongoRepository.java
samleeflang: Hmm it is used by three services, SpecimenService, AnnotationService and DigitalMediaService. Will change to getByVersion

src/test/java/eu/dissco/backend/repository/ElasticSearchRepositoryIT.java
samleeflang: It's not necessary to expose this ports and when you run it while you have an instance of elastic running locally on 9200 it will conflict and be unable to run. Testcontainer normally just picks a random port in the (30XXX range I believe) which are generally always free. That way we don't get unnecessary port conflicts.

src/test/java/eu/dissco/backend/repository/ElasticSearchRepositoryIT.java
samleeflang: Wasn't sure and noticed nothing was happening with the result. But I guess removing a get won't fail the test. Will remove

src/test/java/eu/dissco/backend/repository/SpecimenRepositoryIT.java
samleeflang: No not in this case. I inserted a bunch of records and to make them unique the get a number at the end. So the TestUtils.ID (20.5000.1025/ABC-123-XY3) won't match any of these.

src/main/java/eu/dissco/backend/repository/SpecimenRepository.java
samleeflang: Yes we need to look at it at some point as the same method is also in the ElastiRepository, there is some duplication there. Will revert this here

src/main/java/eu/dissco/backend/service/AnnotationService.java
samleeflang: Will change

src/main/java/eu/dissco/backend/controller/DigitalMediaObjectController.java
samleeflang: Ah yes needed a bit of soul-searching for this one but I remember. It gets the version for the digitalMediaObject. However, all the logic was in the repository and getting the version was removed for it. I made the functionality to get the version generic in the mongodbRepo. I didn't wanted anymore scope creep so to convert the generic setup to JSON:API I created a new userstory. As all other controllers didn't have JSON:API for this yet I decided to remove it here.

src/main/java/eu/dissco/backend/repository/DigitalMediaObjectRepository.java
samleeflang: Hmm no I don't think so, checked and this method uses the
`.where(NEW_DIGITAL_MEDIA_OBJECT.DIGITAL_SPECIMEN_ID.eq(id))` (line 48)
So it gives back all digitalMedia for a specific specimen, think the name is correct.

src/main/java/eu/dissco/backend/configuration/ApplicationConfiguration.java
southeo: Okay, I'll align the handle server with this for now, let me know if anything changes

src/test/java/eu/dissco/backend/repository/ElasticSearchRepositoryIT.java
southeo: Not in the tests themselves, but if you need to debug something later on it might be nice to know which port the testcontainer is running on. Otherwise, it'll just run on a random port. Up to you

southeo:
Looks good to me!

southeo (approved):

PR #32, opened 2023-03-02
Align parsing version to datamodel

samleeflang:
Older version was based on the parsed database records.
It means that new versions will work but older versions in the old datamodel might break.

### DIFF
--- a/src/main/java/eu/dissco/backend/service/AnnotationService.java
+++ b/src/main/java/eu/dissco/backend/service/AnnotationService.java
@@ -119,12 +119,18 @@ public class AnnotationService {
 
   private AnnotationResponse mapResponseToAnnotationResponse(JsonNode response) {
     var annotation = response.get("annotation");
-    return new AnnotationResponse(response.get("id").asText(), response.get("version").asInt(),
-        annotation.get("type").asText(), annotation.get("motivation").asText(),
-        annotation.get("target"), annotation.get("body"), annotation.get("preferenceScore").asInt(),
+    return new AnnotationResponse(response.get("id").asText(),
+        response.get("version").asInt(),
+        annotation.get("type").asText(),
+        annotation.get("motivation").asText(),
+        annotation.get("target"),
+        annotation.get("body"),
+        annotation.get("preferenceScore").asInt(),
         annotation.get("creator").asText(),
-        Instant.ofEpochSecond(annotation.get("created").asLong()), annotation.get("generator"),
-        Instant.ofEpochSecond(annotation.get("generated").asLong()), null);
+        Instant.ofEpochSecond(annotation.get("created").asLong()),
+        annotation.get("generator"),
+        Instant.ofEpochSecond(annotation.get("generated").asLong()),
+        null);
   }
 
   public AnnotationResponse getAnnotationByVersion(String id, int version)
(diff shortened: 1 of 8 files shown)


src/main/java/eu/dissco/backend/service/AnnotationService.java
```diff
@@ -119,12 +119,18 @@ public AnnotationResponse persistAnnotation(AnnotationRequest annotation, String
...
-    return new AnnotationResponse(response.get("id").asText(), response.get("version").asInt(),
-        annotation.get("type").asText(), annotation.get("motivation").asText(),
-        annotation.get("target"), annotation.get("body"), annotation.get("preferenceScore").asInt(),
+    return new AnnotationResponse(response.get("id").asText(),
+        response.get("version").asInt(),
+        annotation.get("type").asText(),
+        annotation.get("motivation").asText(),
+        annotation.get("target"),
+        annotation.get("body"),
+        annotation.get("preferenceScore").asInt(),
```
TomDijkema, line 128: Much better to read

src/main/java/eu/dissco/backend/service/AnnotationService.java
```diff
@@ -119,12 +119,18 @@ public AnnotationResponse persistAnnotation(AnnotationRequest annotation, String
...
         annotation.get("creator").asText(),
-        Instant.ofEpochSecond(annotation.get("created").asLong()), annotation.get("generator"),
-        Instant.ofEpochSecond(annotation.get("generated").asLong()), null);
+        Instant.ofEpochSecond(annotation.get("created").asLong()),
+        annotation.get("generator"),
+        Instant.ofEpochSecond(annotation.get("generated").asLong()),
```
TomDijkema, line 132: Is the transfer from epoch to datetime already on the schedule or is it not relevant to this part?

src/main/java/eu/dissco/backend/service/AnnotationService.java
```diff
@@ -134,19 +140,20 @@ public AnnotationResponse getAnnotationByVersion(String id, int version)
...
+        annotation.get("type").asText(),
+        annotation.get("motivation").asText(),
+        annotation.get("target"),
+        annotation.get("body"),
+        annotation.get("preferenceScore").asInt(),
+        annotation.get("creator").asText(),
+        Instant.ofEpochSecond(annotation.get("created").asInt()),
+        annotation.get("generator"),
+        Instant.ofEpochSecond(annotation.get("generated").asInt()),
+        result.get("deleted") == null ? null : Instant.ofEpochSecond(annotation.get("deleted").asInt())
```
TomDijkema, line 156: I like this approach far more than just result

TomDijkema (approved):

src/main/java/eu/dissco/backend/service/AnnotationService.java
samleeflang: Yes it is on the backlog.

src/main/java/eu/dissco/backend/service/AnnotationService.java
samleeflang: Yes so the main thing is that the metadata is generic for each event object. So we have event (metadata) -> eventRecord (metadata of object (pid, version,created) -> Actual object. Based on the SubjectType (an event metadata field) you can know which object is in the eventRecord. So I think it is more generic and more machine readable. Although fieldnames will probably change when we align with Prov-O

PR #33, opened 2023-03-13
Remove organisation and mapping data flow

southeo:
Removed the following data flows (they have been moved to orchestration service)
- Organisation controller, service, repository
- Mapping Record controller, service, repository

Also removed helper classes used only by these services.


samleeflang:
There seems to some confusion here.
The SourceSystem stuff was transferred right, not the organisation stuff?
The SourceSystem endpoint can go from the backend, but the organisation stuff needs to remain (I think)

PR #35, opened 2023-03-14
Feature/all calls to json api

southeo:
[Jira User story for this PR](https://naturalis.atlassian.net/browse/DD-240?atlOrigin=eyJpIjoiNGIwZDNjZTIxYjExNGRlMGFjOTgzOWQ0MTllZTRlM2YiLCJwIjoiaiJ9 )

"Jsonified" i.e. the process of aligning to the JsonApi standard for both request bodies (where applicable) and response bodies.

Where a single resource is obtained, a JsonApiWrapper is returned. A list of resources returns a JsonApiListResponseWrapper. The difference is that the former only has one JsonApiData object, and the latter has a list of Data objects. If the call is paginated, the JsonApiListResponseWrapper also has a linksObject that includes prev, next, and first. Both objects always include a links object with "self". 

Fully Jsonified
- Annotation
- Digital Media Object
- Specimen

Partially Jsonified:
- Organisation: the function getOrganisationNames() still returns list of strings

Unchanged:
- OrganisationDocument: Need to decide the use case for this endpoint
- SourceSystem/Mapping: Moved to orchestration service
- User: Already done

Related User Stories

- The organisation endpoint is not paginated. [Jira ticket](https://naturalis.atlassian.net/browse/DD-316?atlOrigin=eyJpIjoiZjRhYWJhODk1OTlkNDdmYzlmNDljNzk2ZjgwOTQ0YzciLCJwIjoiaiJ9)
- Exceptions have not been aligned to JsonApi standard. [Jira ticket](https://naturalis.atlassian.net/browse/DD-342?atlOrigin=eyJpIjoiMzZhM2IwN2U0Y2MxNDVjNThmOGQwMWE2NzJiNGM4YmMiLCJwIjoiaiJ9).


src/main/java/eu/dissco/backend/domain/jsonapi/ObjectType.java
```diff
@@ -0,0 +1,19 @@
+package eu.dissco.backend.domain.jsonapi;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public enum ObjectType {
```
samleeflang, line 5: Don't see much usage of this enum, only on one line in AnnotationController. What is the use of this enum?

src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
```diff
@@ -46,14 +46,14 @@ public List<JsonApiData> getAnnotationsForUserJsonResponse(String userId, int pa
...
+  public JsonApiData getAnnotation(String id) {
     return context.select(NEW_ANNOTATION.asterisk())
         .from(NEW_ANNOTATION)
         .where(NEW_ANNOTATION.ID.eq(id))
-        .fetchOne(this::mapToAnnotation);
+        .fetchOne(this::mapToJsonApiData);
   }
 
-  public List<AnnotationResponse> getAnnotations(int pageNumber, int pageSize) {
+  public List<AnnotationResponse> getAnnotationsObject(int pageNumber, int pageSize) {
```
samleeflang, line 56: Is it correct that some still use the 'old' objects and some the new?

src/main/java/eu/dissco/backend/repository/DigitalMediaObjectRepository.java
```diff
@@ -72,6 +58,13 @@ public List<String> getDigitalMediaIdsForSpecimen(String id) {
         .fetch(Record1::value1);
   }
 
+  public List<DigitalMediaObject> getDigitalMediaForSpecimenObject(String id) {
```
samleeflang, line 61: What is the difference between this and `getDigitalMediaForSpecimen` is this method still needed/used

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -37,94 +35,50 @@ public List<DigitalSpecimen> search(String query, int pageNumber, int pageSize)
...
-  }
-
-  public List<AnnotationResponse> getLatestAnnotation() throws IOException {
-    var searchRequest = new SearchRequest.Builder().index(INDEX).q(ANNOTATION_EXISTS_QUERY)
-        .sort(s -> s.field(f -> f.field(FIELD_CREATED).order(SortOrder.Desc))).size(10).build();
-    return client.search(searchRequest, ObjectNode.class).hits().hits().stream().map(Hit::source)
-        .map(this::mapToAnnotationResponse).toList();
+  private JsonApiData mapDigitalSpecimenToJsonApiData(ObjectNode json){
+    var digitalSpecimen = (ObjectNode) json.get("digitalSpecimen");
+    return  new JsonApiData(digitalSpecimen.get("id").asText(), digitalSpecimen.get("type").asText(), digitalSpecimen );
```
samleeflang, line 64: What worries me a bit is that we take the object as-is over. We might want to map to an object first and then put it in attributes, then at least we are sure we serve the same object from database or elastic

src/main/java/eu/dissco/backend/repository/SpecimenRepository.java
```diff
@@ -21,23 +23,57 @@ public class SpecimenRepository {
...
 
-  private DigitalSpecimen mapper(Record dbRecord) {
+  public JsonApiData getLatestSpecimenById(String id){
+    return context.select(NEW_DIGITAL_SPECIMEN.asterisk())
+        .from(NEW_DIGITAL_SPECIMEN)
+        .where(NEW_DIGITAL_SPECIMEN.ID.eq(id))
+        .fetchOne(this::mapToJsonApiData);
+  }
+
+  private JsonApiData mapToJsonApiData(Record dbRecord){
```
samleeflang, line 49: I think it might be better to keep some structure here and first map it to a domain object, so we are sure all objects returned look the same. Then we can add that as the attributes in the JSON:API object

src/main/java/eu/dissco/backend/controller/ControllerUtils.java
```diff
@@ -0,0 +1,11 @@
+package eu.dissco.backend.controller;
+
+public class ControllerUtils {
+  private ControllerUtils(){}
+
+  protected static final String DEFAULT_PAGE_NUM = "1";
+  protected static final String DEFAULT_PAGE_SIZE = "10";
+
+  protected static final String SANDBOX_URI = "https://sandbox.dissco.tech/";
```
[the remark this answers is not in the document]
samleeflang: Additionally can we put these values in a property file with these values as default. If we ever want to change them or the DNS change we can just change the property file

src/main/java/eu/dissco/backend/service/AnnotationService.java
```diff
@@ -33,137 +37,90 @@ public class AnnotationService {
...
+      String userId) {
+    return new AnnotationEvent(annotation.type(), annotation.motivation(), userId, Instant.now(),
+        annotation.target(), annotation.body());
+  }
+
+  public JsonApiWrapper updateAnnotation(String id, AnnotationRequest annotation, String userId, String path)
+      throws NoAnnotationFoundException {
+    var result = repository.getAnnotationForUser(id, userId);
+    if (result > 0) {
+      return persistAnnotation(annotation, userId, path);
```
samleeflang, line 91: Do we have a user story to change this as we need to support partially filled requests?

src/main/java/eu/dissco/backend/service/ServiceUtils.java
```diff
@@ -0,0 +1,21 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.List;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class ServiceUtils {
+  private ServiceUtils(){}
+
+  @Autowired
+  private static ObjectMapper mapper;
```
samleeflang, line 12: Does this work? This is not really a utils functions as it depends upon another class (objectMapper). At some point it might be better to make a superclass for the services. If it works, let's leave it for now.

samleeflang: Tested but it breaks for me.
If you want it to be a utils you would have to pass the mapper as an argument in the method.

src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -36,92 +41,91 @@ public class SpecimenService {
...
-        attributes.get("ods:datasetId").asText(),
-        attributes.get("ods:physicalSpecimenCollection").asText(),
-        attributes.get("ods:sourceSystemId").asText(),
-        attributes,
-        digitalSpecimen.get("ods:originalAttributes"),
-        digitalSpecimen.get("ods:attributes").get("dwca:id").asText()
-    );
+  public JsonApiWrapper getSpecimenByVersion(String id, int version, String path)
+      throws JsonProcessingException, NotFoundException {
+    var specimen = mongoRepository.getByVersion(id, version, "digital_specimen_provenance");
```
samleeflang, line 80: We manipulate the request from MongoDB. We cannot just through the whole JSON from mongo to the user. We will create different objects for elastic, db and mongo. We need to harmonise this data in the domain object and then use this in the JSON:API response

src/test/java/eu/dissco/backend/TestUtils.java
```diff
@@ -34,7 +31,10 @@ public class TestUtils {
   public static final String ID_ALT = PREFIX + "/" + "AAA-111-ZZZ";
   public static final String TARGET_ID = PREFIX + "/TAR_GET_001";
 
-  public static final ObjectMapper MAPPER = new ObjectMapper().findAndRegisterModules();
+  public static final ObjectMapper MAPPER = new ObjectMapper()
+      .findAndRegisterModules()
+      .setDateFormat(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ"))
```
samleeflang, line 36: Did you get this working? It still gave to many SSSSS's for me

samleeflang (commented):
Hi Sou, good to see we are moving towards JSON:API standard.
Couple of things:
- Exception handling -> I don't think we comply to JSON:API there
- Updating -> I don't think we comply to JSON:API there
- Domain objects, I would very much prefer that we first map to a domain object and then put it in the JSON:API wrapper. By pushing things as is we create three different datamodels which is an issue to consumers
- The static class for the versions is not working. Try to functionally/smoke test the changes

Could you also write through the JSON:API and see if we are not missing anything.
For larger things feel free to create a new user story for it.

src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
southeo: Not this function, but the function `getForTargetObject()` is used by the the specimenService and digitalMediaObjectService*. The "old" object is useful because we return a json with values for a specimen/mediaObject plus with its annotations. See [`getDigitalMediaObjectFull()`](https://github.com/DiSSCo/dissco-core-backend/blob/be8faab704fd6f6ffe616d699b56d59bd3cab09b/src/main/java/eu/dissco/backend/service/DigitalMediaObjectService.java#L44)  and [`getSpecimenByIdFull()`](https://github.com/DiSSCo/dissco-core-backend/blob/be8faab704fd6f6ffe616d699b56d59bd3cab09b/src/main/java/eu/dissco/backend/service/SpecimenService.java#L97)

(Well actually the function `getForTargetObject()` is used by the annotationService, which is called by the other services, but that's semantics)

src/main/java/eu/dissco/backend/service/AnnotationService.java
southeo: [Jira](https://naturalis.atlassian.net/browse/DD-390?atlOrigin=eyJpIjoiZGUzYTVhNzI5ZmQxNDRhNWJiMGJhMDczM2E4YjA2NWMiLCJwIjoiaiJ9)

src/main/java/eu/dissco/backend/domain/jsonapi/ObjectType.java
southeo: Removed

src/main/java/eu/dissco/backend/service/SpecimenService.java
southeo: Restored

src/main/java/eu/dissco/backend/service/ServiceUtils.java
southeo: mapper added to function parameter

src/main/java/eu/dissco/backend/controller/ControllerUtils.java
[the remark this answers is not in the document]
southeo: For default page num/page size, I don't think we can, at least not using @Value annotation. @Value injects value after constructor is called. @PathVariables need to be constants; the program won't compile if we set the values this way.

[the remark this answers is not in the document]
southeo: I can move sandboxUri to application.properties, there's an example in annotationController. Not sure if there's a point in using application.properties just for one value, though

src/test/java/eu/dissco/backend/TestUtils.java
southeo: All it took was 3 different de/serializers :)

southeo: technically 2.5

samleeflang: Nice, also looked at a couple of blogs but it seems to also add the nanoseconds in the Jackson Instant Serializer (https://www.javadoc.io/static/com.fasterxml.jackson.datatype/jackson-datatype-jsr310/2.6.5/index.html?com/fasterxml/jackson/datatype/jsr310/ser/InstantSerializer.html), strange... But this seems to work nicely :+1:

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
samleeflang: This has been fixed

samleeflang (dismissed):
One minor comment nothing blocking merging

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -90,9 +93,22 @@ private DigitalSpecimen mapToDigitalSpecimen(ObjectNode json) {
   }
 
   private Instant parseDate(JsonNode instantNode) {
```
samleeflang, line 95: Sorry this will change in my PR... There is an item on the backlog already to fix the datatime format of elastic... driving me nuts at the moment. 
https://naturalis.atlassian.net/browse/DD-382

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -31,6 +32,8 @@ public class ElasticSearchRepository {
   private final ElasticsearchClient client;
   private final DateTimeFormatter formatter;
 
+  private final Pattern pattern = Pattern.compile("-?\\d+(\\.\\d+)?");
```
samleeflang, line 35: Stupid question but where is the pattern used?

samleeflang (approved):
One question about the pattern but I think I removed it already.
Let's merge it and see what happens with the PR behind it.

PR #36, opened 2023-03-28
Feature/generic specimen search

samleeflang:
Adds a generic search endpoint:
- https://naturalis.atlassian.net/browse/DD-215
- https://naturalis.atlassian.net/browse/DD-394
- https://naturalis.atlassian.net/browse/DD-396
Also change `organization` to `organisation`


src/main/java/eu/dissco/backend/client/AnnotationClient.java
```diff
@@ -15,7 +15,7 @@ public interface AnnotationClient {
   @PostMapping(value = "")
   JsonNode postAnnotation(@RequestBody AnnotationEvent event);
 
-  @DeleteMapping(value = "/{prefix}/{postfix}")
+  @DeleteMapping(value = "/{prefix}/{suffix}")
   void deleteAnnotation(@PathVariable("prefix") String prefix,
-      @PathVariable("postfix") String postfix);
+      @PathVariable("suffix") String suffix);
```
TomDijkema, line 20: Nice!

src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -43,39 +45,62 @@ public class SpecimenService {
...
+    map.put("license", "digitalSpecimen.ods:attributes.dcterms:license.keyword");
+    map.put("hasMedia", "digitalSpecimen.ods:attributes.ods:hasMedia.keyword");
+    map.put("organisationId", "digitalSpecimen.ods:attributes.ods:organisationId.keyword");
+    map.put("organisationName", "digitalSpecimen.ods:attributes.ods:organisationName.keyword");
+    map.put("sourceSystemId", "digitalSpecimen.ods:attributes.ods:sourceSystemId.keyword");
+    map.put("type", "digitalSpecimen.ods:type.keyword");
+    map.put("specimenName", "digitalSpecimen.ods:attributes.ods:specimenName");
+    map.put("q", "q");
+    return map;
+  }
```
TomDijkema, line 72: The mapping is some extra work, but I think this will make it much cleaner to work with the API

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -24,33 +26,46 @@
...
+          query = new Query.Builder().queryString(q -> q.query(sanitisedValue)).build();
+        } else {
+          query = new Query.Builder().term(t -> t.field(entry.getKey()).value(value)).build();
+        }
+        queries.add(query);
+      }
+    }
+    return queries;
+  }
+
```
TomDijkema, line 65: So, this is where you build the query for ElasticSearch from the given params right?

samleeflang: Yes, so we use the boolean queries from elastic.
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html
We use the should query.
Each term is a should and for which one value should match.
So we got a bunch of shoulds and the `minimumShouldMatch` indicates how many shoulds shoud match

TomDijkema: Yes, so thanks to the should clause it can fetch multiple records right? With must it would to be focused on a certain specimen and the filter clause is something in between I guess, it says is not dependent on the score of the search? (Don't really get that one).

samleeflang: Yes so we always get back an array. Even for physicalSpecimenId which should always return 1 record. The score, what I read this will be used. So the more the should hit the higher the score. This is mainly important with for the free text as it could partly match. Most are exact matches which is either true or false.
For documentation:
`The bool query takes a more-matches-is-better approach, so the score from each matching must or should clause will be added together to provide the final _score for each document.`

src/main/java/eu/dissco/backend/domain/jsonapi/JsonApiLinksFull.java
```diff
@@ -18,11 +20,45 @@ public JsonApiLinksFull(int pageNum, int pageSize, boolean hasNext, String path)
...
+    }
+    var pathFull = builder.toString();
+
+    this.self = buildPathPages(pathFull, pageNum, pageSize, isFirst);
+    this.first = buildPathPages(pathFull, 1, pageSize, isFirst);
+    this.prev = (pageNum <= 1) ? null : buildPathPages(pathFull, (pageNum - 1), pageSize, isFirst);
+    this.next = (hasNext) ? buildPathPages(pathFull, (pageNum + 1), pageSize, isFirst) : null;
+  }
+
+  private String buildPathPages(String baseString, int pageNum, int pageSize, boolean isFirst) {
```
southeo, line 54: Nice method here, can you reuse it in the other constructor? (Line 14)

src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -167,42 +187,76 @@ private JsonNode generateIdNode() {
...
+        digitalSpecimen.get("ods:attributes").get("dwca:id").asText()
+    );
   }
 
-  private JsonApiListResponseWrapper wrapListResponse(List<DigitalSpecimen> digitalSpecimenList, int pageSize, int pageNumber, String path){
-    List<JsonApiData> dataNodePlusOne = new ArrayList<>();
-    digitalSpecimenList.forEach(specimen -> dataNodePlusOne.add(new JsonApiData(specimen.id(), specimen.type(), specimen, mapper)));
+  private JsonApiListResponseWrapper wrapListResponse(List<DigitalSpecimen> digitalSpecimenList,
+      int pageSize, int pageNumber, String path) {
+    var dataNodePlusOne = digitalSpecimenList.stream().map(specimen -> new JsonApiData(specimen.id(), specimen.type(), specimen, mapper)).toList();
```
southeo, line 214: Why change `.forEach(...)` to `.stream.map(...)` ? Just curious if there's performance differences

src/test/java/eu/dissco/backend/repository/ElasticSearchRepositoryIT.java
```diff
@@ -91,83 +85,95 @@ static void initContainer() {
...
+    return Stream.of(
+        Arguments.of("digitalSpecimen.ods:physicalSpecimenId.keyword", "global_id_45634"),
+        Arguments.of("q", PREFIX + "/0")
+    );
   }
 
-  @AfterAll
-  public static void closeResources() throws Exception {
-    restClient.close();
+  @BeforeEach
```
southeo, line 100: Is there a reason you reordered these? I think @BeforeAll, @BeforeEach, @AfterAll is clearer.

src/test/java/eu/dissco/backend/repository/ElasticSearchRepositoryIT.java
```diff
@@ -91,83 +85,95 @@ static void initContainer() {
...
 
     // When
-    var responseReceived = repository.search(targetId, 1, 1);
+    var responseReceived = repository.search(Map.of(field, List.of(value)), 1, 1);
 
     // Then
     assertThat(responseReceived).contains(targetSpecimen);
   }
 
+  @AfterEach
```
southeo, line 127: Should move this to the top with the rest of the setup/destroy functions

southeo (changes requested):
Kudos to removing postfix/organization! :partying_face: 

Just a few things

src/main/java/eu/dissco/backend/domain/jsonapi/JsonApiLinksFull.java
samleeflang: I don't really like the MultiValue map why would you make this special map why you could have a Map<X, List<X>> doesn't make sense to me... But Spring requires the MultiValue map so we need that to get it parsed in but would rather look something clearer for the next bit.

src/test/java/eu/dissco/backend/repository/ElasticSearchRepositoryIT.java
samleeflang: Yes agree, will change

src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -167,42 +187,76 @@ private JsonNode generateIdNode() {
...
+    return wrapListResponse(specimenPlusOne, pageNumber, pageSize, params, path);
+  }
+
+  private static int getIntParam(String paramName, MultiValueMap<String, String> params, String defaultValue) {
+    var paramValue = params.getOrDefault(paramName, List.of(defaultValue));
+    if (paramValue.size() > 1){
+      log.warn("Taking first value for param: {} values: {}", paramName, paramValue);
+    }
+    try{
+      return Integer.parseInt(paramValue.get(0));
```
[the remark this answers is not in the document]
samleeflang: Good catch... phfff all this for just a pageNumber

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -61,7 +76,8 @@ public List<AnnotationResponse> getLatestAnnotations(int pageNumber, int pageSiz
       throws IOException {
     var offset = getOffset(pageNumber, pageSize);
 
-    var searchRequest = new SearchRequest.Builder().index(INDEX).q(ANNOTATION_EXISTS_QUERY)
+    var searchRequest = new SearchRequest.Builder()
+        .index(ANNOTATION_INDEX)
         .sort(s -> s.field(f -> f.field(FIELD_CREATED).order(SortOrder.Desc))).from(offset)
         .size(pageSize).build();
     return client.search(searchRequest, ObjectNode.class).hits().hits().stream().map(Hit::source)
```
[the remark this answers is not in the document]
samleeflang: I would say no. But we don't really have delete functionality for specimen data. For annotations I noticed that we do archive the annotation and handles but don't remove it from elastic. I created an item for it: https://naturalis.atlassian.net/browse/DD-414

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -24,33 +26,46 @@
...
+  private static final String ANNOTATION_INDEX = "annotation";
   private static final String FIELD_CREATED = "created";
   private static final String FIELD_GENERATED = "generated";
   private final ElasticsearchClient client;
-  private final DateTimeFormatter formatter;
-
-  private final Pattern pattern = Pattern.compile("-?\\d+(\\.\\d+)?");
 
-  public List<DigitalSpecimen> search(String query, int pageNumber, int pageSize)
+  public List<DigitalSpecimen> search(Map<String, List<String>> params, int pageNumber,
```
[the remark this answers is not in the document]
samleeflang: https://naturalis.atlassian.net/browse/DD-414

src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -167,42 +187,76 @@ private JsonNode generateIdNode() {
...
+  }
 
+  private Map<String, List<String>> mapParams(MultiValueMap<String, String> params) {
+    var mappedParams = new HashMap<String, List<String>>();
+    for (var entry : params.entrySet()) {
+      var mappedParam = paramMapping.get(entry.getKey());
+      if (mappedParam != null) {
+        mappedParams.put(mappedParam, entry.getValue());
+      } else {
+        log.warn("Did not find mapping for key: {}", entry.getKey());
```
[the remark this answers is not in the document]
samleeflang: Ok yes I wasn't sure on the one hand we want to do our best to at least provide for something, on the other hand we don't want to provide incorrect data. But think you are right and it makes more sense to return a 4XX something message

src/main/java/eu/dissco/backend/service/SpecimenService.java
samleeflang: There is nothing wrong with the forEach but I don't really like that we init a list and then loop over another to fill it. This is bit more concise and just returns a new list of the values we want. I often try to make methods to return the object instead of munipulate a parameter. So instead of `void something(List<X> x)` and in the method add things I think it is more readable to have `List<X> something()`.

southeo (approved):
:partying_face:

PR #37, opened 2023-04-03
Add aggregation endpoint

samleeflang:
Adds a aggregation endpoint which returns the format.
Example query:
`localhost:8080/api/v1/specimens/aggregation?country=Finland&sourceSystemId=20.5000.1025/5N1-FST-J0E`
Returns the following format.
```{
    "data": {
        "id": "1145151679",
        "type": "aggregations",
        "attributes": {
            "midsLevel": {
                "0": 9
            },
            "license": {
                "CC0 1.0": 9
            },
            "country": {
                "Finland": 9
            },
            "organisationName": {},
            "sourceSystemId": {
                "20.5000.1025/5N1-FST-J0E": 9
            },
            "typeStatus": {},
            "datasetId": {},
            "hasMedia": {
                "false": 5,
                "true": 4
            }
        }
    },
    "links": {
        "self": "https://sandbox.dissco.tech/api/v1/specimens/aggregation?country=Finland&sourceSystemId=20.5000.1025/5N1-FST-J0E"
    }
}
```

User story:
https://naturalis.atlassian.net/browse/DD-395

### DIFF
--- a/src/main/java/eu/dissco/backend/domain/MappingTerms.java
+++ b/src/main/java/eu/dissco/backend/domain/MappingTerms.java
@@ -1,0 +1,70 @@
+package eu.dissco.backend.domain;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public enum MappingTerms {
+  COUNTRY("country", "digitalSpecimen.ods:attributes.dwc:country.keyword"),
+  COUNTRY_CODE("countryCode", "digitalSpecimen.ods:attributes.dwc:countryCode.keyword"),
+  MIDS_LEVEL("midsLevel", "midsLevel"),
+  PHYSICAL_SPECIMEN_ID("physicalSpecimenId", "digitalSpecimen.ods:physicalSpecimenId.keyword"),
+  TYPE_STATUS("typeStatus", "digitalSpecimen.ods:attributes.dwc:typeStatus.keyword"),
+  LICENSE("license", "digitalSpecimen.ods:attributes.dcterms:license.keyword"),
+  HAS_MEDIA("hasMedia", "digitalSpecimen.ods:attributes.ods:hasMedia.keyword"),
+  ORGANISATION_ID("organisationId", "digitalSpecimen.ods:attributes.ods:organisationId.keyword"),
+  ORGANISATION_NAME("organisationName",
+      "digitalSpecimen.ods:attributes.ods:organisationName.keyword"),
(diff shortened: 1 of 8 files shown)


src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -267,7 +254,7 @@ private Map<String, List<String>> mapParams(MultiValueMap<String, String> params
       throws UnknownParameterException {
     var mappedParams = new HashMap<String, List<String>>();
     for (var entry : params.entrySet()) {
-      var mappedParam = paramMapping.get(entry.getKey());
+      var mappedParam = MappingTerms.getParamMapping().get(entry.getKey());
```
southeo, line 257: You can remove `MappingTerms` and just statically import this function

src/main/java/eu/dissco/backend/domain/MappingTerms.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private final String name;
+  private final String fullName;
+
+  MappingTerms(String name, String fullName) {
+    this.name = name;
+    this.fullName = fullName;
+  }
+
+  public static Map<String, String> getParamMapping() {
```
southeo, line 32: I don't think these should be functions. It doesn't make sense to re-construct the list/hash map on every call. Would be better to make these both static & final variables instead.

src/main/java/eu/dissco/backend/service/SpecimenService.java
southeo: Also, you can simplify this function. Instead of the function `mapParams` returning the whole hashmap, you could pass `entry.getKey()` as an argument and have` mapParams` just return the value.

src/main/java/eu/dissco/backend/domain/MappingTerms.java
```diff
@@ -0,0 +1,70 @@
+package eu.dissco.backend.domain;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public enum MappingTerms {
```
southeo, line 8: I don't see why this needs to be an enum and not a regular utility class

src/main/java/eu/dissco/backend/service/SpecimenService.java
southeo: Also might be nice to make this an `Optional`

TomDijkema (dismissed):
Looks good, I don't have particular comments

southeo (approved):
:sunglasses:

PR #39, opened 2023-04-11
Fix timestamp + add version to annotation

samleeflang:
Timestamp was missing timezone.
Also added version in the annotation response as it was missing.


southeo (approved):
Yep

PR #41, opened 2023-04-19
Feature/spring boot 3.0

southeo:
[Jira](https://naturalis.atlassian.net/browse/DD-243?atlOrigin=eyJpIjoiMmY5M2E5M2I2MGMzNDg2Y2FlMTQ3MGM4YTcxMGZmYTkiLCJwIjoiaiJ9)


pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>2.7.8</version>
+    <version>3.0.5</version>
```
samleeflang, line 9: 3.0.6 should help with at least on of the above issues

pom.xml
```diff
@@ -119,6 +108,15 @@
       <artifactId>mongodb-driver</artifactId>
       <version>${mongodb-driver.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
```
samleeflang, line 112: Why is this one needed?

src/main/java/eu/dissco/backend/security/JwtAuthConverter.java
```diff
@@ -0,0 +1,37 @@
+package eu.dissco.backend.security;
+
+import java.util.Collection;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.jetbrains.annotations.NotNull;
```
samleeflang, line 7: Don't think this is the correct one

samleeflang (commented):
Couple of comments

PR #44, opened 2023-05-15
Get full specimen by version

samleeflang:
Add endpoint to get full specimen data by id and version.

https://naturalis.atlassian.net/browse/DD-497


southeo (approved):
Looks good!

PR #46, opened 2023-05-17
Create endpoint for searching within term

samleeflang:
Endpoint where we can de prefix searches for term.
So when you put in the term and the search value it will check all aggregated values for that term.
For example you have Spain, Final and sans origin for term Country.
If you put in country and S as value it will return Spain and sand origin with their respective values.
If you put in country and Sp it will only return Spain.

Add hits = 0 as it only needs to return the buckets, not the search results, is probably quicker.

https://naturalis.atlassian.net/browse/DD-491

### DIFF
--- a/src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
+++ b/src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
@@ -204,10 +204,21 @@ public class ElasticSearchRepository {
         .trackTotalHits(t -> t.enabled(Boolean.TRUE))
         .aggregations(mappingTerm.getName(),
             AggregationBuilders.terms().field(mappingTerm.getFullName()).build()._toAggregation())
+        .size(0)
         .build();
     var aggregation = client.search(aggregationRequest, ObjectNode.class);
     var totalRecords = aggregation.hits().total().value();
     var aggregationResult = collectResult(aggregation.aggregations());
     return Pair.of(totalRecords, aggregationResult);
+  }
+
+  public Map<String, Map<String, Long>> searchTermValue(String name, String field, String value) throws IOException {
+    var searchQuery = new SearchRequest.Builder().index(DIGITAL_SPECIMEN_INDEX)
+        .query( q -> q.prefix( m -> m.field(field).value(value)))
+        .aggregations(name, agg -> agg.terms(t -> t.field(field)))
+        .size(0)
+        .build();
+    var aggregation = client.search(searchQuery, ObjectNode.class);
+    return collectResult(aggregation.aggregations());
   }
 }
--- a/src/main/java/eu/dissco/backend/service/SpecimenService.java
+++ b/src/main/java/eu/dissco/backend/service/SpecimenService.java
@@ -23,7 +23,6 @@ import eu.dissco.backend.exceptions.NotFoundException;
 import eu.dissco.backend.exceptions.UnknownParameterException;
 import eu.dissco.backend.repository.ElasticSearchRepository;
 import eu.dissco.backend.repository.MongoRepository;
-import eu.dissco.backend.repository.RepositoryUtils;
 import eu.dissco.backend.repository.SpecimenRepository;
 import java.io.IOException;
 import java.time.Instant;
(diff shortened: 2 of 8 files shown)


src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -287,18 +287,30 @@ private Map<String, List<String>> mapParams(MultiValueMap<String, String> params
...
   public JsonApiWrapper discipline(String path) throws IOException {
     var disciplineResult = elasticRepository.getAggregation(TOPIC_DISCIPLINE);
-    var dataNode = new JsonApiData(String.valueOf(path.hashCode()), "aggregations",
+    var dataNode = new JsonApiData(String.valueOf(path.hashCode()), AGGREGATIONS_TYPE,
         mapper.valueToTree(disciplineResult.getRight()));
     return new JsonApiWrapper(dataNode,
         new JsonApiLinks(path), new JsonApiMeta(disciplineResult.getLeft()));
   }
 
+  public JsonApiWrapper searchTermValue(String name, String value, String path)
```
southeo, line 304: "name" here means term name? In the controller this variable is called "term". That's more clear to me, but I see that "name" is an attribute of Term, so i see why it's called this.

src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -287,18 +287,30 @@ private Map<String, List<String>> mapParams(MultiValueMap<String, String> params
...
+      throws UnknownParameterException, IOException {
+    var mappedTerm = getMappedTerm(name);
+    if (mappedTerm.isPresent()){
+      var aggregations = elasticRepository.searchTermValue(name, mappedTerm.get(), value);
```
southeo, line 308: Wouldn't name and mappedTerm.get() always be the same?

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -204,10 +204,21 @@ public Pair<Long, Map<String, Map<String, Long>>> getAggregation(MappingTerms ma
...
             AggregationBuilders.terms().field(mappingTerm.getFullName()).build()._toAggregation())
+        .size(0)
         .build();
     var aggregation = client.search(aggregationRequest, ObjectNode.class);
     var totalRecords = aggregation.hits().total().value();
     var aggregationResult = collectResult(aggregation.aggregations());
     return Pair.of(totalRecords, aggregationResult);
   }
+
+  public Map<String, Map<String, Long>> searchTermValue(String name, String field, String value) throws IOException {
```
southeo, line 215: What's the difference between name and field

southeo (commented):
So to clarify: every time a user types a character, it will trigger a call to this endpoint?

samleeflang:
Yes each time a letter is typed it could make this call. It looks pretty fast, 30-60 ms when I tested it

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
samleeflang: So the name would be `country` and the field would be `digitalSpecimen.ods:attributes.dwc:country.keyword`

src/main/java/eu/dissco/backend/service/SpecimenService.java
samleeflang: No, the getMappedTerm get's the fullName/elasticName for the field. In the MappingTerms there is a map with the name and fullName. In a few cases they are the same, for example `midsLevel`

src/main/java/eu/dissco/backend/service/SpecimenService.java
samleeflang: Yes tricky with all the different names. For the frontend it is the term, but for elastic it is the name... So frontend would like to keep it term. But here it is not really the full term but the (simplified) name of the term. Does that make sense?

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
samleeflang: We set it here so that the bucketName is equal to the original name we got from the frontend instead of the full elastic name

southeo (approved):
Thanks for clarifying!

PR #47, opened 2023-05-19
JsonApi-compliant exceptions

southeo:
[JsonApi Standard](https://jsonapi.org/examples/#error-objects-basics)

JsonApi wants us to return a list of error objects, so our response looks like:

{
  "errors": [
    {
      "status": "404",
      "title":  "Not found",
      "detail": "Requested resource {} not found"
    }
  ]
}

It's a lot of nesting, considering we only return one error object at a time. :upside_down_face: It is what it is       

[Jira](https://naturalis.atlassian.net/browse/DD-342?atlOrigin=eyJpIjoiY2FhZTVkMjI2MzhjNGMzZTgwZjU4MzNmYjZhOWM5NDciLCJwIjoiaiJ9)


src/main/java/eu/dissco/backend/controller/RestResponseEntityExceptionHandler.java
```diff
@@ -18,44 +19,79 @@ public class RestResponseEntityExceptionHandler extends ResponseEntityExceptionH
...
   }
 
   @ResponseStatus(HttpStatus.CONFLICT)
   @ExceptionHandler(ConflictException.class)
-  public ResponseEntity<Void> handleException(ConflictException e) {
-    return ResponseEntity.status(HttpStatus.CONFLICT).build();
+  public ResponseEntity<ExceptionResponseWrapper> handleException(ConflictException e) {
+    var exceptionResponse =  new ExceptionResponseWrapper(
+            HttpStatus.CONFLICT,
+            "Conflict",
```
samleeflang, line 36: Maybe a bit more context. Think this is a conflict exception that only occurs when the id in the object doesn't match the id in the url. So maybe a "Conflicting id's in message" might cover it better

samleeflang: And the message here might be null I think

src/main/java/eu/dissco/backend/controller/RestResponseEntityExceptionHandler.java
```diff
@@ -18,44 +19,79 @@ public class RestResponseEntityExceptionHandler extends ResponseEntityExceptionH
...
 
   @ResponseStatus(HttpStatus.NOT_FOUND)
   @ExceptionHandler(NotFoundException.class)
-  public ResponseEntity<Void> handleException(NotFoundException e) {
-    return ResponseEntity.status(HttpStatus.NOT_FOUND).build();
+  public ResponseEntity<ExceptionResponseWrapper> handleException(NotFoundException e) {
+    var exceptionResponse =  new ExceptionResponseWrapper(
+            HttpStatus.NOT_FOUND,
+            "Resource Not Found",
+            e.getMessage()
```
samleeflang, line 48: This might also be empty, maybe check if we add a message to all exception (the ones we throw ourselves)

samleeflang (commented):
See if you can make the message a bit more specific for the user. I also think some exception messages are empty

src/main/java/eu/dissco/backend/controller/RestResponseEntityExceptionHandler.java
southeo: Added more detailed message. Right now, the conflicting ids are captured in the logs but not in an exception. Is that something we want to provide in an error object?

southeo:
Added default message for the exceptions that can have null messages.

src/main/java/eu/dissco/backend/controller/RestResponseEntityExceptionHandler.java
southeo: This exception is parent to 3 exceptions: InvalidIdException, InvalidTypeException, UserExistsException. However, we handle the 3 exceptions at the parent level, so the granularity is lost, and we don't capture any information in a message. Do we need 3 inheritors of this exception?

samleeflang: Ok yes, we could do it this way. Or remove the non-arg constructor from the exceptions, forcing it everywhere to have a message for the exception. The granulatiry could then be in the message that is included in the exception.

samleeflang (approved):

PR #50, opened 2023-05-31
Fix offset bug

samleeflang:
Adding 1 to the pageSize to check if there is a next page happens too early.
This created an issue when calculating the offset as it user pageSize +1 to calculate offset.
That should be fixed, impacted more than excepted.

https://naturalis.atlassian.net/browse/DD-517

### DIFF
--- a/src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
+++ b/src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
@@ -1,6 +1,7 @@
 package eu.dissco.backend.repository;
 
 import static eu.dissco.backend.database.jooq.Tables.NEW_ANNOTATION;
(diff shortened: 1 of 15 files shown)


src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
```diff
@@ -29,11 +30,12 @@ public AnnotationResponse getAnnotation(String id) {
   }
 
   public List<AnnotationResponse> getAnnotations(int pageNumber, int pageSize) {
-    int offset = getOffset(pageNumber, pageSize);
+    var offset = getOffset(pageNumber, pageSize);
+    var pageSizePlusOne = pageSize + ONE_TO_CHECK_NEXT;
```
TomDijkema, line 34: Simple but good solution :+1:

TomDijkema (approved):
It is in the little things ;)

PR #56, opened 2023-08-10
Specify json creator

southeo:
Jackson didn't know which of the 2 constructors to use for the request type. Specified this, now it should be groovy.


TomDijkema (approved):
Sonar perfection

PR #57, opened 2023-08-24
add count in meta field

southeo:
Include total count of hits for specimens and annotations by a given creator. Value is stored in meta field. 

Jira: 

https://naturalis.atlassian.net/browse/DD-619?atlOrigin=eyJpIjoiYWE4NTc3NWU2NjI5NDFiYTg3MTI3ZDkyNGQ2ODk5NzAiLCJwIjoiaiJ9

https://naturalis.atlassian.net/browse/DD-620?atlOrigin=eyJpIjoiYjUxNDA1OGMxMmY1NGM5MDhjMjcxN2E1MDNhNjU1ODYiLCJwIjoiaiJ9

### DIFF
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -1,10 +1,15 @@
 # Date: June 16, 2023
 # Issue: snappy-java high issue, https://www.cvedetails.com/cve/CVE-2023-34453/
 # Solution: Spring needs to update its version of kafka
-CVE-2023-34455
+# CVE-2023-34455
 
 # Date: August 24, 2023
 # Issue: snappy-java used by kafka, which is used by spring-kafka
 # Solution: Spring needs to update its version of kafka
-CVE-2023-34453
-CVE-2023-34454
+# CVE-2023-34453
+# CVE-2023-34454
+
+# Date: August 25
+# Manually set kafka version to 3.5.1. This version addresses the above CVEs
+# Should remove this property once spring-kafka updates
+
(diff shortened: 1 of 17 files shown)


src/main/java/eu/dissco/backend/service/SpecimenService.java
```diff
@@ -64,15 +64,16 @@ public class SpecimenService {
   private final AnnotationService annotationService;
   private final MongoRepository mongoRepository;
 
-  public JsonApiListResponseWrapper getSpecimen(int pageNumber, int pageSize, String path) {
-    var digitalSpecimenList = repository.getSpecimensLatest(pageNumber, pageSize);
-    return wrapListResponse(digitalSpecimenList, pageSize, pageNumber, path);
+  public JsonApiListResponseWrapper getSpecimen(int pageNumber, int pageSize, String path)
```
samleeflang, line 67: Wondering, are the db methods still used?

.github/workflows/.trivyignore
```diff
@@ -1,4 +1,10 @@
 # Date: June 16, 2023
 # Issue: snappy-java high issue, https://www.cvedetails.com/cve/CVE-2023-34453/
 # Solution: Spring needs to update its version
-CVE-2023-34455
\ No newline at end of file
+CVE-2023-34455
```
samleeflang, line 4: Is this still relevant or can we upgrade spring?

src/main/java/eu/dissco/backend/service/SpecimenService.java
southeo: Nope. Thought they were, but only in testing...methods (and tests) are removed

.github/workflows/.trivyignore
southeo: We use latest version of spring-kafka, which uses 3.4 kafka by default, but according to [this](https://spring.io/projects/spring-kafka), kafka 3.5 should be compatible with spring. 

Manually upgraded kafka in the meantime, made a note in trivyignore.

PR #61, opened 2023-10-13
Feature/enforce orcid

southeo:
Enforce orcid on annotation creation

Assumes that user exists for updating and deleting purposes.

### DIFF
--- a/src/main/java/eu/dissco/backend/service/UserService.java
+++ b/src/main/java/eu/dissco/backend/service/UserService.java
@@ -8,7 +8,6 @@ import eu.dissco.backend.domain.jsonapi.JsonApiWrapper;
 import eu.dissco.backend.exceptions.ConflictException;
 import eu.dissco.backend.exceptions.InvalidIdException;
 import eu.dissco.backend.exceptions.InvalidTypeException;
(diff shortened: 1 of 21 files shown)


src/main/java/eu/dissco/backend/service/UserService.java
```diff
@@ -47,39 +46,27 @@ private void checkType(JsonApiWrapper request) throws InvalidTypeException {
...
   }
 
-  public JsonApiData findUser(String id) throws NotFoundException {
-    var optionalUser = repository.find(id);
-    if (optionalUser.isPresent()) {
-      return new JsonApiData(id, TYPE, mapper.valueToTree(optionalUser.get()));
-    } else {
-      throw new NotFoundException();
-    }
+  public JsonApiData findUser(String id) {
```
samleeflang, line 49: Oeh yes. I think I added the notFoundException because it is part of JSON:API. Could you check if you are still aligned with the specs, I think that was the reason it was working this way

samleeflang (commented):
I believe I added the NotFoundException and the optional stuff because we need to be JSON:API compliant. Could you check if we aren't breaking compliancy by removing it?

samleeflang (approved):
:flower_playing_cards:

PR #62, opened 2023-10-17
annotations model

southeo:
- Remove separate annotationRequest/Event/Response. it's all one object: Annotation
- Backend adds "dcterms:created" timestamp (on create operation only) and "oa:creator" to Annotations object before sending it to processing server
- Update db tables
- Needs to be merged in tandem with changes in the processor 

Sonar is mad that I'm not testing all conditions on the Motivation.java `fromString()` function. Sonar and I will just have to agree disagree

known issues: I want to apply a jsonschema validation check on the request to make sure only expected fields are coming in. User story is already up

Related PR - [Annotation Processor](https://github.com/DiSSCo/dissco-core-annotation-processing-service/pull/18)


src/main/java/eu/dissco/backend/controller/AnnotationController.java
```diff
@@ -97,7 +97,7 @@ public ResponseEntity<JsonApiWrapper> createAnnotation(Authentication authentica
     var annotation = getAnnotationFromRequest(requestBody);
     var userId = getNameFromToken(authentication);
     log.info("Received new annotation from user: {}", userId);
-    var annotationResponse = service.persistAnnotation(annotation, userId, getPath(request));
+    var annotationResponse = service.persistAnnotation(annotation, userId, getPath(request), false);
```
samleeflang, line 100: Why is this false needed?

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -81,35 +88,43 @@ public Pair<Long, List<DigitalSpecimenWrapper>> getLatestSpecimen(int pageNumber
...
-        annotation.get("target"),
-        annotation.get("body"),
-        annotation.get("preferenceScore").asInt(),
-        getText(annotation, "creator"),
-        createdOn,
-        annotation.get("generator"),
-        generatedOn,
-        null);
+  private Annotation mapToAnnotationResponse(ObjectNode annotation) {
+    var createdOn = parseDate(annotation.get(FIELD_CREATED_ANNOTATION));
```
samleeflang, line 104: Wonder if we cannot just map with the objectMapper directly to Annotation

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -140,18 +155,18 @@ private Instant parseDate(JsonNode instantNode) {
     return null;
   }
 
-  private String getText(JsonNode digitalSpecimen, String element) {
-    var jsonNode = digitalSpecimen.get(element);
+  private String getText(JsonNode annotation, String element) {
```
samleeflang, line 158: Do we use this function genericly? Maybe just call it jsonNode or node.

src/main/java/eu/dissco/backend/service/AnnotationService.java
```diff
@@ -86,50 +77,63 @@ public JsonApiListResponseWrapper getAnnotations(int pageNumber, int pageSize,
     return wrapListResponse(annotationsPlusOne, pageNumber, pageSize, path);
   }
 
-  public JsonApiWrapper persistAnnotation(AnnotationRequest annotationRequest, String userId,
-      String path) throws ForbiddenException {
+  public JsonApiWrapper persistAnnotation(Annotation annotationRequest, String userId,
+      String path, boolean isUpdate) throws ForbiddenException, JsonProcessingException {
```
samleeflang, line 81: I think `isUpdate` can be removed

src/main/java/eu/dissco/backend/client/AnnotationClient.java
```diff
@@ -13,7 +14,10 @@
 public interface AnnotationClient {
 
   @PostMapping(value = "")
-  JsonNode postAnnotation(@RequestBody AnnotationEvent event);
+  JsonNode postAnnotation(@RequestBody Annotation annotation);
+
+  @PatchMapping(value = "")
```
samleeflang, line 19: As indicated in the processor PR, would be good to put to id in the url as the delete

src/test/java/eu/dissco/backend/service/AnnotationServiceTest.java
```diff
@@ -105,7 +105,7 @@ void testGetAnnotationsForUser() throws Exception {
     var receivedResponse = service.getAnnotationsForUser(userId, pageNumber, pageSize, path);
```
samleeflang, line 105: Maybe also remove response from `receivedResponse`?

src/test/java/eu/dissco/backend/service/AnnotationServiceTest.java
```diff
@@ -205,96 +205,83 @@ void testGetAnnotationsJsonResponseLastPage() {
...
   void testGetLatestAnnotations() throws IOException {
     int pageSize = 15;
     int pageNumber = 1;
     String path = SANDBOX_URI + "api/v1/annotations/latest/json";
-    var expectedResponse = givenAnnotationJsonResponse(path, pageNumber, pageSize,
+    var expected = givenAnnotationJsonResponse(path, pageNumber, pageSize,
         USER_ID_TOKEN, ID, true);
-    var elasticResponse = Collections.nCopies(pageSize + 1, givenAnnotationResponse());
+    var elasticResponse = Collections.nCopies(pageSize + 1,
+        AnnotationUtils.givenAnnotationResponse(ID));
```
samleeflang, line 219: AnnotationUtils can be a static import

samleeflang (commented):
Very happy that this is almost through. Just one more `new` table to go :wink:

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
southeo: done

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
southeo: removed altogether

src/test/java/eu/dissco/backend/service/AnnotationServiceTest.java
southeo: not sure what you mean here

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
samleeflang: Sweet that saves a lot of code

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
samleeflang: Even better

src/test/java/eu/dissco/backend/service/AnnotationServiceTest.java
samleeflang: You remove `Response` from `expectedResponse` only change I see in this unit test. However, the received value is still called `receivedResponse`. Not sure why this change was needed but thought it would like better if it `Response` was removed consistently

samleeflang (dismissed):
Looks good! Reacted on one open point but that is not blocking.

samleeflang (approved):

PR #63, opened 2023-10-24
add handle function

southeo:
Adds handle classes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,6 +60,10 @@
     </dependency>
     <dependency>
       <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-webflux</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-validation</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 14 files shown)


src/main/java/eu/dissco/backend/web/HandleComponent.java
```diff
@@ -0,0 +1,139 @@
...
+
+  public void rollbackHandleUpdate(List<JsonNode> request)
+      throws PidCreationException, PidAuthenticationException {
+    log.info("Rolling back handle update");
+    var requestBody = BodyInserters.fromValue(request);
+    var response = sendRequest(HttpMethod.DELETE, requestBody, "rollback/update");
+    getFutureResponse(response);
+  }
+
+  public void rollbackFromPhysId(List<String> physIds) {
```
samleeflang, line 65: There is stuff in here we don't need right?

src/test/java/eu/dissco/backend/utils/HandleUtils.java
```diff
@@ -0,0 +1,80 @@
...
+          "data": {
+            "type": "digitalSpecimen",
+            "attributes": {
+              "fdoProfile": "https://hdl.handle.net/21.T11148/d8de0819e144e4096645",
+              "digitalObjectType": "https://hdl.handle.net/21.T11148/894b1e6cad57e921764e",
+              "issuedForAgent": "https://ror.org/0566bfb96",
+              "primarySpecimenObjectId": "https://geocollections.info/specimen/23602",
+              "normalisedPrimarySpecimenObjectId":"https://geocollections.info/specimen/23602",
+              "primarySpecimenObjectIdType": "Global",
+              "specimenHost": "https://ror.org/0443cwa12",
```
samleeflang, line 22: Doesn't look like appropriate FDO requests.

pom.xml
```diff
@@ -58,6 +58,10 @@
       <artifactId>snappy-java</artifactId>
       <version>${snappy-java.version}</version>
     </dependency>
+    <dependency>
```
samleeflang, line 61: Ok yes I guess for the API call to the annotation processing service we use a different client (openfeign I believe). But that will be removed at some point and then this will be the only client.

samleeflang (commented):
There might be a couple things that can be removed as we won't do specimen. Also most examples/mock data is aimed at specimen

PR #69, opened 2023-11-06
Enforce new target vocabulary

southeo:
Target Type can only be:

- DigitalSpecimen
- Annotation
- MediaObject


southeo:
We will use PIDs instead of a set vocabulary

PR #70, opened 2023-11-07
Add mjr endpoint

southeo:
- Get MJR record by job id
- Get paginated list of MJRs by creator (may filter by state (failed, scheduled, completed)

[JIRA](https://naturalis.atlassian.net/browse/DD-810?atlOrigin=eyJpIjoiZjExYTQ4OWRkMTUxNDZlZWIxMDhlN2M1N2I2NzQzOGIiLCJwIjoiaiJ9)

### DIFF
--- a/src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
+++ b/src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
@@ -1,0 +1,56 @@
+package eu.dissco.backend.controller;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import eu.dissco.backend.domain.AnnotationState;
+import eu.dissco.backend.domain.jsonapi.JsonApiListResponseWrapper;
+import eu.dissco.backend.domain.jsonapi.JsonApiWrapper;
+import eu.dissco.backend.exceptions.NotFoundException;
+import eu.dissco.backend.properties.ApplicationProperties;
+import eu.dissco.backend.service.MasJobRecordService;
+import jakarta.servlet.http.HttpServletRequest;
+import java.util.UUID;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.RestControllerAdvice;
+
+@Slf4j
+@RestControllerAdvice
+@RestController
+@RequestMapping("api/v1/mjr")
+public class MasJobRecordController extends BaseController {
+
+  private final MasJobRecordService service;
+
+  public MasJobRecordController(ObjectMapper mapper,
+      ApplicationProperties applicationProperties, MasJobRecordService service) {
+    super(mapper, applicationProperties);
+    this.service = service;
+  }
+
+  @GetMapping(value = "/{jobId}", produces = MediaType.APPLICATION_JSON_VALUE)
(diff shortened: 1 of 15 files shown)


src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.RestControllerAdvice;
+
+@Slf4j
+@RestControllerAdvice
+@RestController
+@RequestMapping("api/v1/mjr")
```
samleeflang, line 25: The call I am missing is to get jobs called for a particular specimen.
Maybe this should be a call/operation on the specimen (specimens/XXX/XXX/mjr) returns all mjr's for that specimen. Maybe with a request param where you can add the status, so you can select only the ones that are scheduled or failed.
Probably best to make a new task for this.

src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
```diff
@@ -0,0 +1,56 @@
...
+    this.service = service;
+  }
+
+  @GetMapping(value = "/{jobId}", produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<JsonApiWrapper> getMasJobRecord(
+      @PathVariable("jobId") UUID jobId, HttpServletRequest request) throws NotFoundException {
+    return ResponseEntity.ok().body(service.getMasJobRecordById(jobId, getPath(request)));
+  }
+
+  @GetMapping(value = "/creator/"
```
samleeflang, line 42: Wonder if this is a call/operation on the user object and should therefore be in the user controller. What do you think?

src/main/java/eu/dissco/backend/service/MasJobRecordService.java
```diff
@@ -16,15 +25,60 @@
...
+    if (state == null) {
+      masJobRecordsPlusOne = masJobRecordRepository.getMasJobRecordsByCreator(creatorId, pageNum,
+          pageSizeToCheckNext);
+    } else {
+      masJobRecordsPlusOne = masJobRecordRepository.getMasJobRecordsByCreatorAndStatus(creatorId,
+          state.getState(), pageNum, pageSizeToCheckNext);
+    }
+    boolean hasNext = masJobRecordsPlusOne.size() > pageSize;
+    var masJobRecords = hasNext ? masJobRecordsPlusOne.subList(0, pageSize) : masJobRecordsPlusOne;
+    List<JsonApiData> dataList = new ArrayList<>();
```
samleeflang, line 56: Couldn't this be a stream?

samleeflang (commented):
Nice addition! Looking forward seeing this in the FE

src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
southeo: [Added new task](https://naturalis.atlassian.net/browse/DD-834)

src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
southeo: "Creator" here is the MAS Id, not user that called the user that scheduled the MAS, so I think this endpoint is the best. 
We don't actually capture the user information in the Mas Job Record. I can a user might want to see all their scheduled jobs. Added a [new jira story](https://naturalis.atlassian.net/browse/DD-835) to modify this object

samleeflang: Oh, yes sorry that would be nice, let's add that info :+1:

samleeflang (dismissed):
:clap:

samleeflang (approved):

PR #73, opened 2023-11-15
Feature/json schema validation for annotations

southeo:
Schema for incoming web annotations. 
Required:
- "rdf:type",
 - "oa:motivation",
 - "oa:target",
 - "oa:body"

Optional: 
- oa:motivatedBy

Backend adds the "creator" and "created" fields to the annotation then passes it to the processor


samleeflang (approved):
Sweet, nice addition. Nothing to complain :fireworks:

PR #76, opened 2023-11-24
Orcid endpoint

southeo:
Given an orcid, return user information

[DD-797](https://naturalis.atlassian.net/browse/DD-797?atlOrigin=eyJpIjoiOWJlOTdkOTdhODBiNGVhMGIzMTZjYTcyZDFkYjhkZGUiLCJwIjoiaiJ9)

### DIFF
--- a/src/main/java/eu/dissco/backend/controller/UserController.java
+++ b/src/main/java/eu/dissco/backend/controller/UserController.java
@@ -71,6 +71,16 @@ public class UserController extends BaseController {
     return ResponseEntity.ok(new JsonApiWrapper(response, new JsonApiLinks(getPath(request))));
   }
 
+  @ResponseStatus(HttpStatus.OK)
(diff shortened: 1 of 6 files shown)


src/main/java/eu/dissco/backend/controller/UserController.java
```diff
@@ -71,6 +71,16 @@ public ResponseEntity<JsonApiWrapper> getUser(Authentication authentication,
     return ResponseEntity.ok(new JsonApiWrapper(response, new JsonApiLinks(getPath(request))));
   }
 
+  @ResponseStatus(HttpStatus.OK)
+  @GetMapping(value = "/{orcid}/orcid")
```
samleeflang, line 75: I believe the standard is to put the path varaible at the end so /orcid/{orcid{

samleeflang (commented):
Just one comment on the path

samleeflang (approved):
:+1:

PR #84, opened 2023-12-04
Feature/running mas endpoint

southeo:
Adds endpoint for MAS to mark a given job as `running`

Upgrade to spring 3.2.0

### DIFF
--- a/src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
+++ b/src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
@@ -13,6 +13,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PatchMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
@@ -53,4 +54,11 @@ public class MasJobRecordController extends BaseController {
 
   }
 
+  @PatchMapping(value = "/{creatorId}/{masJobId}/running")
+  public ResponseEntity<Void> markMjrAsRunning(
+      @PathVariable("creatorId") String creatorId,
+      @PathVariable("masJobId") UUID masJobId) throws NotFoundException {
+    service.markMasJobRecordAsRunning(creatorId, masJobId);
+    return ResponseEntity.ok().build();
+  }
 }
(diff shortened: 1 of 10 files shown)


src/main/java/eu/dissco/backend/repository/MasJobRecordRepository.java
```diff
@@ -101,6 +102,17 @@ public void markMasJobRecordsAsFailed(List<UUID> ids) {
         .execute();
   }
 
+  public void markMasJobRecordAsRunning(String creatorId, UUID masJobId) throws NotFoundException {
+    var updated = context.update(MAS_JOB_RECORD)
+        .set(MAS_JOB_RECORD.STATE, AnnotationState.RUNNING.getState())
+        .where(MAS_JOB_RECORD.JOB_ID.eq(masJobId))
+        .and(MAS_JOB_RECORD.CREATOR_ID.eq(creatorId))
+        .execute();
+    if (updated == 0){
```
samleeflang, line 111: I would probably do this in the service and only keep the query in the repository

src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
```diff
@@ -53,4 +54,11 @@ public ResponseEntity<JsonApiListResponseWrapper> getMasJobRecordsForCreator(
 
   }
 
+  @PatchMapping(value = "/{creatorId}/{masJobId}/running")
```
samleeflang, line 57: We discussed this. Let's leave it this way for now, maybe in the future we can remove the creator (if each MAS get's it's own client token or something).

src/main/java/eu/dissco/backend/repository/MasJobRecordRepository.java
```diff
@@ -101,6 +102,17 @@ public void markMasJobRecordsAsFailed(List<UUID> ids) {
         .execute();
   }
 
+  public void markMasJobRecordAsRunning(String creatorId, UUID masJobId) throws NotFoundException {
+    var updated = context.update(MAS_JOB_RECORD)
+        .set(MAS_JOB_RECORD.STATE, AnnotationState.RUNNING.getState())
```
samleeflang, line 107: At some point it would be nice to have these in db enums as well. But not real prio, just saves space and make sure it fails if something else goes into that column https://www.postgresql.org/docs/current/datatype-enum.html

samleeflang (commented):
One small comment and two remarks

src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
southeo: noted in something to discuss further

src/main/java/eu/dissco/backend/repository/MasJobRecordRepository.java
southeo: made a jira story

samleeflang (approved):
:flatbread:

PR #90, opened 2023-12-18
Split creator id

southeo:
- Split creator id in /running endpoint
- Change to GET request to allow securityfilterchain to permit this function without authentication
- Only SCHEDULED jobs can be set to RUNNING


TomDijkema (approved):
:sunflower:

PR #91, opened 2023-12-18
MAS Job Record Data Model Changes

southeo:
- Adds handle client and token authenticator client
- All handle requests will be the same, so no need for a dedicated service. I generate the request from application properties, but maybe it would make more sense to just read a file where the request could be? 
- JobId (UUID) -> JobHandle (String)
- Adds db enum for field JOB_STATE
- Adds target_type in the mas job record table. Distinguishing DIGITAL_SPECIMEN and MEDIA_OBJECT lets the front end link to the appropriate resource
- Refactoring some terms
      - creatorId -> masId (in db: creator_id -> mas_id)
      - AnnotationState / state - > jobState (in db: state -> job_state)
   

Notes
- This PR interfaces with the table mas_job_record_new. Before merging, we'll want to replace mas_job_record table with the new version of the table (and update the annotation processing service as well).


src/main/java/eu/dissco/backend/component/FdoComponent.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Getter;
+import org.springframework.stereotype.Component;
+
+@Component
+@Getter
+public class FdoComponent {
+
+  private final JsonNode postRequest;
+
+  public FdoComponent(ObjectMapper mapper, FdoProperties fdoProperties) {
```
samleeflang, line 15: As this is a Component, you should be able to just inject the mapper

src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
```diff
@@ -45,21 +47,24 @@ public ResponseEntity<JsonApiListResponseWrapper> getMasJobRecordsForCreator(
...
       HttpServletRequest request) {
     return ResponseEntity.ok().body(
-        service.getMasJobRecordsByCreator(creatorId, getPath(request), pageNumber, pageSize,
+        service.getMasJobRecordsByMasId(creatorId, getPath(request), pageNumber, pageSize,
             state));
 
   }
 
-  @GetMapping(value = "/{creatorIdPrefix}/{creatorIdSuffix}/{masJobId}/running")
+  @GetMapping(value = "/{masIdPrefix}/{masIdSuffix}/{jobIdPrefix}/{jobIdSuffix}/running")
```
samleeflang, line 58: This won't look nice (20.1000.10/XXX-XXX-XXX/20.1000.10/YYY-YYY-YYY) but don't think we can do much about that

src/main/java/eu/dissco/backend/properties/FdoProperties.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Data;
+import lombok.RequiredArgsConstructor;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+
+@Validated
+@Data
+@RequiredArgsConstructor
+@ConfigurationProperties("fdo")
```
samleeflang, line 13: Should we specify here that this is about the mjr?

src/main/resources/jooq-configuration.xml
```diff
@@ -18,7 +18,7 @@
         annotation |
         user |
         machine_annotation_services |
-        mas_job_record
+        mas_job_record_new
```
samleeflang, line 21: Can we do this without the new. Something like:
- first code review
- PR approved
- Fix table to new layout
- Rerun JOOQ and make last fixes
- Merge
- Deploy
Downtime shouldn't be that long and it is just test for acc we could do it right the first time as this hasn't been to acc yet.

samleeflang (commented):
Just a couple small things. It would be nice if we don't get stuck with another `new` in a table name. So I would suggest to get this approved and the quickly change the table and the Jooq stuff and then quickly merge. That way we have little downtime and arent' stuck with the `new`.
Also remember to deploy it and add all the new required env's to the deployment files.

src/test/resources/db/migration/V1__init_db.sql
```diff
@@ -110,25 +110,17 @@ CREATE TABLE machine_annotation_services
...
 (
-    job_id         uuid default uuid_generate_v4() not null
-        constraint mas_job_record_pk
+    job_id         text                     not null
+        constraint mas_job_record_new_pk
             primary key,
-    state          text                            not null,
-    creator_id     text                            not null,
-    time_started   timestamp with time zone        not null,
+    job_state      text check ( job_state in ('SCHEDULED', 'FAILED', 'COMPLETED', 'RUNNING') )            not null,
```
samleeflang, line 118: Just one more comment. Types in postgres can be done with: https://www.postgresql.org/docs/current/datatype-enum.html

src/main/java/eu/dissco/backend/component/FdoComponent.java
southeo: I did it this way because otherwise we have to set the post request in a @postConstruct, which is hard to test (would need autowiring/integration testing). I think we also said we wanted to avoid postConstructors in general ([this article ](https://levelup.gitconnected.com/stop-using-postconstruct-in-your-java-applications-2a66fb202cb8)discusses issues with unit testing & postconstruct, among other issues)

src/main/java/eu/dissco/backend/controller/MasJobRecordController.java
southeo: I know :( 

Eventually, we will want to have MAS profiles for authorization, so that will reduce the endpoint complexity.

src/main/java/eu/dissco/backend/properties/FdoProperties.java
southeo: I don't think that's necessary. Currently the only objects managed by the backend that receive a handle are MJRs, but  might change. We've come close to linking the handle api with the backend a few times now, and we might decide more objects need handles down the line. I'd leave it for now.

src/test/resources/db/migration/V1__init_db.sql
southeo: Oof this is a doozy...so I tried this and I don't think that's possible with Jooq's enum converter. 

Given this enum converter: 

```
<configuration>
 <generator>
   <database>
     <forcedTypes>
       <forcedType>
         <userType>eu.dissco.backend.domain.MasJobState</userType>
         <enumConverter>true</enumConverter>
         <includeExpression>.*\.JOB_STATE</includeExpression>
       </forcedType>
     </forcedTypes>
   </database>
 </generator>
</configuration>
```

What this mapping does is map the enum MasJobState to a varchar, which is sent to the database. **It does not create a custom postgres type that can be used in a java application.** 

I did create a postgres type called JobState, and when I tried it out (both in test and using the actual table), the database rejected the data because jooq was mapping the value of the enum to a varchar and not JobState.

src/main/resources/jooq-configuration.xml
southeo: Sounds good, really didn't like making the `new_` table.

src/main/java/eu/dissco/backend/component/FdoComponent.java
samleeflang: I don't think so. This is a Component class which means it is a Bean and has access to other Beans. The ObjectMapper is also a Bean and can just be injected here as a private final variable as we do in other places. So we wouldn't need to use a PostConstruct method.

src/test/resources/db/migration/V1__init_db.sql
samleeflang: It should be possible to do. Now we don't get the advantages of the enum in the database. If you cannot forceType it you could just map it. When you run JOOQ with the enum (without forcedTypes) it should generate Enum classes for you. You would then need to map to these classes (just as the JSONB column).

samleeflang (commented):
Two or three comments still open. Let me know if you want to discuss these

src/main/java/eu/dissco/backend/component/FdoComponent.java
southeo: The postRequest is not a bean. We could inject the mapper/FdoProperties, but that only works if we use the @RequiredArgsConstructor, so we wouldnt be initializing the postRequest JsonNode during construction. I could make postRequest a bean, or I could remove the @component flag, or I could read a static json from a properties file.

samleeflang: Ok yes let's go with this.

samleeflang (approved):
:accept:

samleeflang (approved):

PR #92, opened 2023-12-19
Handle null annotation node

southeo:
Catch NPE

### DIFF
--- a/src/main/java/eu/dissco/backend/repository/MasJobRecordRepository.java
+++ b/src/main/java/eu/dissco/backend/repository/MasJobRecordRepository.java
@@ -40,7 +40,7 @@ public class MasJobRecordRepository {
       int pageNum, int pageSize) {
     var offset = getOffset(pageNum, pageSize);
     var condition = MAS_JOB_RECORD.TARGET_ID.eq(targetId);
-    if (state != null){
+    if (state != null) {
       condition = condition.and(MAS_JOB_RECORD.STATE.eq(state.getState()));
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/eu/dissco/backend/repository/MasJobRecordRepository.java
```diff
@@ -123,6 +123,9 @@ private Record5<String, String, String, Instant, String> mjrToRecord(MasJobRecor
 
   private MasJobRecordFull recordToMasJobRecord(Record dbRecord) {
     try {
+      var dataNode = dbRecord.get(MAS_JOB_RECORD.ANNOTATIONS) != null ?
+          mapper.readValue(dbRecord.get(MAS_JOB_RECORD.ANNOTATIONS).data(), JsonNode.class) :
+          null;
       return new MasJobRecordFull(
```
southeo, line 129: This function takes the contents pulled from the database into a Java Object. 

`mapper.readValue(dbRecord.get(MAS_JOB_RECORD.ANNOTATIONS).data(), JsonNode.class)` -> takes the value stored in the ANNOTATIONS column in the MAS_JOB_RECORD table and maps it to a JSON Object (which is part of the Java Object MasJobRecordFull). 

Before, if the value of ANNOTATIONS col was null, a null pointer exception (NPE) would be thrown. If this column is null, we automatically set the MasJobRecordField, without triggering the NPE.

src/test/java/eu/dissco/backend/repository/MasJobRecordRepositoryIT.java
```diff
@@ -54,6 +56,19 @@ void testGetMasJobRecordById() throws JsonProcessingException {
     assertThat(result).isPresent().contains(expected);
   }
 
+  @Test
```
southeo, line 59: For full test coverage.

TomDijkema (approved):
:bug:

PR #95, opened 2024-01-15
Return MJR instead of MAS

southeo:
When scheduling a MAS, returns list of scheduled jobs instead of MAS metadata. 

Fixed bug where MjrService was attempting to create handles for 0 records if no MAS records comply to target object. 
[
Jira](https://naturalis.atlassian.net/browse/DD-906)


src/main/java/eu/dissco/backend/service/MasJobRecordService.java
```diff
@@ -101,7 +101,7 @@ public Map<String, String> createMasJobRecord(Set<MachineAnnotationServiceRecord
             orcid))
         .toList();
     masJobRecordRepository.createNewMasJobRecord(masJobRecordList);
-    return masJobRecordList.stream().collect(Collectors.toMap(MasJobRecord::masId, MasJobRecord::jobId));
+    return masJobRecordList.stream().collect(Collectors.toMap(MasJobRecord::masId, mjr -> mjr));
```
samleeflang, line 104: I think `mjr -> mjr` could be `Function.identity()`. Reading this post: https://stackoverflow.com/questions/28032827/java-8-lambdas-function-identity-or-t-t
Could be either way do what you think is best, although I think I mostly used `Function.identity()`

src/test/java/eu/dissco/backend/utils/MachineAnnotationServiceUtils.java
```diff
@@ -38,6 +39,15 @@ public static JsonApiListResponseWrapper givenMasResponse(
     return new JsonApiListResponseWrapper(masRecords, links, new JsonApiMeta(masRecords.size()));
   }
 
+  public static JsonApiListResponseWrapper givenScheduledMasResponse(
+      MasJobRecord masJobRecord,
+      String path) {
```
samleeflang, line 44: Formatting looks a bit off

samleeflang (dismissed):
Nice change

src/main/java/eu/dissco/backend/service/MasJobRecordService.java
southeo: oh I like that, thanks!

samleeflang (approved):

PR #96, opened 2024-01-16
MAS and MJR changes

southeo:
MAS table: batching_permitted
MJR table: batching_requested

Scheduling MAS job accepts additional optional parameter: `batching`. If `batching` is not set, defaults to false. If a user requests batching and the MAS doesn't allow batching, an exception is thrown, and the job is not scheduled, 

Once all PRs are approved, will update machine_annotation_services /mas_job_record table and have this code point to the correct table, not _tmp versions.
- [Orchestration Service](https://github.com/DiSSCo/dissco-core-orchestration-backend/pull/34)

Sonar isn't happy with an 8 parameter function. I could package a few values and unpack them upstream, but that would increase the cognitive complexity rather than decrease it.


src/main/java/eu/dissco/backend/controller/DigitalMediaObjectController.java
```diff
@@ -125,14 +126,16 @@ public ResponseEntity<JsonApiListResponseWrapper> getMasJobRecordForMedia(
   @PostMapping(value = "/{prefix}/{suffix}/mas", produces = MediaType.APPLICATION_JSON_VALUE)
   public ResponseEntity<JsonApiListResponseWrapper> scheduleMassForDigitalMediaObject(
       @PathVariable("prefix") String prefix, @PathVariable("suffix") String suffix,
+      @RequestParam Optional<Boolean> batching,
```
samleeflang, line 129: I think you can set a Default value for this:
`@RequestParam(defaultValue = false) boolean batching`

src/main/java/eu/dissco/backend/service/MachineAnnotationServiceService.java
```diff
@@ -103,6 +108,21 @@ public JsonApiListResponseWrapper scheduleMass(JsonNode flattenObjectData, List<
         new JsonApiMeta(scheduledMasRecords.size()));
   }
 
+  private void validateBatchingRequest(Boolean batchingRequested,
+      List<MachineAnnotationServiceRecord> masRecords)
```
samleeflang, line 112: Think formatting is off here

src/main/java/eu/dissco/backend/service/MachineAnnotationServiceService.java
```diff
@@ -103,6 +108,21 @@ public JsonApiListResponseWrapper scheduleMass(JsonNode flattenObjectData, List<
...
+      throws ConflictException {
+    if (Boolean.FALSE.equals(batchingRequested)) {
```
samleeflang, line 114: This could have been `!batchingRequested` right? Why did you choose for this way?

src/test/java/eu/dissco/backend/service/MachineAnnotationServiceServiceTest.java
```diff
@@ -98,39 +100,51 @@ void testGetMassForObjectNoFilterMatch(List<Pair<String, List<String>>> filters)
...
 
   @Test
-  void testScheduleMass() throws JsonProcessingException {
+  void testScheduleMass() throws JsonProcessingException, ConflictException {
     // Given
     var digitalSpecimen = givenDigitalSpecimenWrapper(ID);
     var masRecord = givenMasRecord(givenFiltersDigitalSpecimen());
     given(repository.getMasRecords(List.of(ID))).willReturn(List.of(masRecord));
-    given(masJobRecordService.createMasJobRecord(Set.of(masRecord), ID, ORCID, MjrTargetType.DIGITAL_SPECIMEN)).willReturn(
+    given(masJobRecordService.createMasJobRecord(Set.of(masRecord), ID, ORCID, MjrTargetType.DIGITAL_SPECIMEN, false)).willReturn(
```
samleeflang, line 108: Might have missed it but is there a test which has batching on `true`?

src/main/java/eu/dissco/backend/controller/DigitalMediaObjectController.java
southeo: I think sonar suggested doing it this way instead (or maybe it was the IDE...?) Can't trigger the recommendation again, so I'll set a defaultValue instead.

src/main/java/eu/dissco/backend/service/MachineAnnotationServiceService.java
southeo: Sonar recommends doing it this way to avoid NPEs

src/test/java/eu/dissco/backend/service/MachineAnnotationServiceServiceTest.java
southeo: line 122: `testScheduleMassInvalidBatchRequest()`
Except for a validation check, batching being true/false doesn't affect the flow in this service (more the masJobRecord service)

samleeflang (approved):

samleeflang (approved):

PR #103, opened 2024-03-14
Don't let trivy break main pipeline

southeo:
Only apply task "Trivy - Stop on Severe Vulnerabilities" during PR event.


samleeflang (approved):
:+1:  only a dozen more project to go through but should be hard if this works

PR #107, opened 2024-04-04
Switch to generic job_state

samleeflang:
Switched from the specific mjr_job_state to a generic job_state type which is also used by the translator job record


samleeflang:
Reverted database change. Let's first get your PR through and then I will build these changes on top of it.

PR #111, opened 2024-04-29
Feature/use builder paradigm

southeo:
Uses @builder on annotation objects to clean up code a bit.


samleeflang (approved):
Sweet! :pie: Sorry I didn't spot this in the initial PR

PR #112, opened 2024-05-08
set rdf:type

southeo:
Lombok `@Builder.default` not behaving as expected and not setting `rdf:type = Annotation`. Making this field final instead worked.


samleeflang (approved):
:saxophone:

PR #113, opened 2024-05-23
align with handle api changes

southeo:
Align with handle api changes: type is PID, remove params the handle api doesn't accept anymore.
Forgot the backend also has a handle component specifically for masJobRecords


samleeflang (approved):
:jellyfish:

PR #114, opened 2024-05-24
add batchId

southeo:
https://github.com/DiSSCo/dissco-core-annotation-processing-service/pull/45

Adds batchId to annotation object

non-breaking change (for once...)


samleeflang (approved):
:shell:

PR #115, opened 2024-05-28
Feature/web batch annotations

southeo:
Only available to those with specific role

GET /batch 
- takes annotation target type (PID for specimen or media type) and search params
- makes elasticsearch query based on params, returns number of objects that would be affected

POST /batch 
- Sends batch request to /batch web endpoint of processing service
- should be done after GET /batch but is not enforced

Also added a BAD_GATEWAY response entity for IOExceptions


src/main/java/eu/dissco/backend/security/WebSecurityConfig.java
```diff
@@ -23,6 +23,8 @@ public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Excepti
     http.authorizeHttpRequests(authorizeHttpRequests -> authorizeHttpRequests
         .requestMatchers(EndpointRequest.to(HealthEndpoint.class)).permitAll()
         .requestMatchers(HttpMethod.GET, "**").permitAll()
+        .requestMatchers(HttpMethod.POST, "*/v1/annotations/batch").hasRole("dissco-web-batch-annotations")
```
samleeflang, line 26: I like the annotations above the endpoint better, as it is clearly visible. You can just in the Controller do a `@hasRole("dissco-web-batch-annotations")` and it will do the same thing. I would say for general things such as all Get endpoints are open we put them here but for roles for specific endpoints we put them on top of the method. What do you think?

samleeflang (commented):
Looks great, couple things, some of which we might tackle in a V2 version

src/main/java/eu/dissco/backend/component/SchemaValidatorComponent.java
```diff
@@ -24,17 +25,34 @@ public void validateAnnotationRequest(Annotation annotation, Boolean isNew)
...
-    log.error("Invalid annotation request. Errors {}", errors);
+    log.error("Invalid annotations request. Errors {}", errors);
     throw new InvalidAnnotationRequestException(errors.toString());
   }
 
+  public void validateAnnotationEventRequest(AnnotationEvent event, boolean isNew)
+      throws InvalidAnnotationRequestException {
+    if (event.annotations().size() != 1 || event.batchMetadata().size() != 1
+        || event.batchMetadata().get(0).getSearchParams().isEmpty()) {
+      var searchParamSize = event.batchMetadata().isEmpty() ? "0"
```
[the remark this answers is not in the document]
southeo: changed to int. think i was doing some string concatenation earlier. it's just for logging so not a big deal but good to be consistent

src/main/java/eu/dissco/backend/repository/ElasticSearchRepository.java
```diff
@@ -262,4 +267,38 @@ public Map<String, Map<String, Long>> aggregateTermValue(String name, String fie
...
+        .build();
+    return client
+        .count(countRequest)
+        .count();
+  }
+
+  private List<Query> generateBatchAnnotationQuery(BatchMetadata batchMetadata) {
+    var qList = new ArrayList<Query>();
+    for (var searchParam : batchMetadata.getSearchParams()) {
+      var key = searchParam.inputField().replaceAll("\\[[^]]*]", "") + ".keyword";
```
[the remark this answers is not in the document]
southeo: ooof you're right...that's a bit of a pain

[Jira](https://naturalis.atlassian.net/browse/DD-1199?atlOrigin=eyJpIjoiNmQzMTM2NTY4OTI5NGUzNmIxYTNlODJhYTA0NmFjZTYiLCJwIjoiaiJ9)

src/main/java/eu/dissco/backend/controller/RestResponseEntityExceptionHandler.java
```diff
@@ -33,7 +33,7 @@ public ResponseEntity<ExceptionResponseWrapper> handleJsonException(JsonProcessi
   @ResponseStatus(HttpStatus.BAD_GATEWAY)
   @ExceptionHandler(IOException.class)
   public ResponseEntity<ExceptionResponseWrapper> handleIOException(java.io.IOException e){
-    logger.error("An IOException has occurred. Check elastic connection.", e);
+    logger.error("An IOException has occurred.", e);
     var exceptionResponse = new ExceptionResponseWrapper(
         HttpStatus.BAD_GATEWAY,
         "ElasticSearch exception",
```
samleeflang, line 39: Sorry my main comment was about this one

southeo: fixed

samleeflang (approved):
:parachute:

PR #116, opened 2024-06-06
Batching Conflict exception should be distinct

southeo:
BatchingNotPermittedException thrown now instead of generic ConflictException when user requests batching on a service that doesn't permit it. 

[Jira](https://naturalis.atlassian.net/browse/DD-1191?atlOrigin=eyJpIjoiYTM4MmY3NDk4YTlmNGY0MDkyYThjMWI2MjVhY2JkZTYiLCJwIjoiaiJ9)


samleeflang (approved):
:tropical_fish:

PR #124, opened 2024-08-06
Bug fix/fix batch annotations

southeo:
Fixes
- Assigns web batch requests job information in the masJobRecord table. The processing service expects batch results to have an entry in this table. May want to rename this to "scheduledJobRecords" down the line
- media object -> digital media

### DIFF
--- a/src/main/java/eu/dissco/backend/domain/annotation/batch/BatchMetadata.java
+++ b/src/main/java/eu/dissco/backend/domain/annotation/batch/BatchMetadata.java
@@ -7,7 +7,7 @@ import lombok.Value;
 
 @Value
 public class BatchMetadata {
-  static Integer placeInBatch = 1;
+  Integer placeInBatch = 1;
   @JsonProperty("searchParams")
   List<SearchParam> searchParams;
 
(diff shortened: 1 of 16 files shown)


src/main/java/eu/dissco/backend/domain/annotation/batch/BatchMetadata.java
```diff
@@ -7,7 +7,7 @@
 
 @Value
 public class BatchMetadata {
-  static Integer placeInBatch = 1;
+  Integer placeInBatch = 1;
```
TomDijkema, line 10: What does this change exactly?

src/main/java/eu/dissco/backend/service/MasJobRecordService.java
```diff
@@ -119,6 +121,22 @@ public Map<String, MasJobRecord> createMasJobRecord(
         .collect(Collectors.toMap(MasJobRecord::masId, Function.identity()));
   }
 
+  public String createJobRecordForDisscover(Annotation annotation, String orcid) {
```
TomDijkema, line 124: Nice name ;)

TomDijkema (approved):
:canada:

src/main/java/eu/dissco/backend/domain/annotation/batch/BatchMetadata.java
southeo: removing `static` means that it searlizes into a json, otherwise the term is ignored when batchMetadata is serialized. We send batchMetadata as a json to the specimen processor, and it's an important part of the pipeline.

PR #128, opened 2024-09-02
remove user table

southeo:
Remove user table, all info is now in token. 
We take `given_name family_name` to make the value for schema:name.

the MJR table still calls the user that scheduled the mas "user_id" instead of "orcid", but it now contains the orcid.

Condition coverage doesn't seem necessary for this one, the conditions are what "fullName" is for the user if the first name or last name is missing...


src/main/java/eu/dissco/backend/controller/BaseController.java
```diff
@@ -26,6 +35,34 @@ public abstract class BaseController {
   protected final ObjectMapper mapper;
   private final ApplicationProperties applicationProperties;
 
+  protected User getUser(Authentication authentication) throws ForbiddenException {
+    var tokenEncoded = ((Jwt) authentication.getPrincipal()).getTokenValue();
+    var tokenBody = tokenEncoded.split("\\.")[1];
+    var tokenString = Base64.getUrlDecoder().decode(tokenBody);
```
samleeflang, line 41: hmm ok, you probably looked at it but there was no easier way to do it? They aren't properties in a kind of Map or thing in the token...

src/main/java/eu/dissco/backend/controller/BaseController.java
```diff
@@ -26,6 +35,34 @@ public abstract class BaseController {
...
+      token = mapper.readTree(tokenString);
+    } catch (IOException e) {
+      log.error("Unable to read authentication token");
+      throw new ForbiddenException("Unable to read authentication token");
+    }
+    if (token.get("orcid") != null) {
+      StringBuilder fullName = new StringBuilder();
+      if (token.get("given_name") != null){
+        fullName.append(token.get("given_name").asText())
+            .append(" ");
```
samleeflang, line 53: Not sure about the append here, if family name is not there it might be an issue. Maybe trim the name before adding it could solve it, or check if full_name is not empty when adding the family name

samleeflang (commented):
Love how much code we can remove. If this is gone, we can also start removing the table. Just two small comments/questions

src/main/java/eu/dissco/backend/controller/BaseController.java
southeo: there's actually a way easier way to do this, fixed now

PR #129, opened 2024-09-04
Add ods:sourceSystemName

samleeflang:
Small PR to add SourceSystem in the aggregations
- updated to latest openDS (otherwise we got errors in the frontend because it got a unrecognised property)
- update of pom

### DIFF
--- a/src/main/java/eu/dissco/backend/domain/DefaultMappingTerms.java
+++ b/src/main/java/eu/dissco/backend/domain/DefaultMappingTerms.java
@@ -18,6 +18,7 @@ public enum DefaultMappingTerms implements MappingTerm {
   ORGANISATION_ID("organisationID", "ods:organisationID.keyword"),
   ORGANISATION_NAME("organisationName", "ods:organisationName.keyword"),
   SOURCE_SYSTEM_ID("sourceSystemID", "ods:sourceSystemID.keyword"),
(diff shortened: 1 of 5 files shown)


TomDijkema (approved):

src/main/java/eu/dissco/backend/domain/DefaultMappingTerms.java
```diff
@@ -57,7 +58,7 @@ private static Set<MappingTerm> fillAggregationList() {
     aggregationTerms.add(HAS_MEDIA);
     aggregationTerms.add(ORGANISATION_NAME);
     aggregationTerms.add(ORGANISATION_ID);
-    aggregationTerms.add(SOURCE_SYSTEM_ID);
+    aggregationTerms.add(SOURCE_SYSTEM_NAME);
```
southeo, line 61: We could keep source system id as an aggregation term?

southeo (approved):
:lab_coat:

src/main/java/eu/dissco/backend/domain/DefaultMappingTerms.java
samleeflang: We could but it doesn't say a lot to the user which is why we wanted to change it. You can still filter on it though so search?sourceSystemID=XXX/XXX would still work. But the list of aggregations would otherwise get a bit long (and eventuelly slow)

TomDijkema (approved):

PR #130, opened 2024-09-09
Include annotation and agent in archive request

southeo:
Sends Annotation and Tombstoning Agent to annotation processing service on archive request
- We need to update the annotation data in the processing service to include tombstone metadata -> we need the resolved annotation to do this
- We need the tombstoning agent in the tombstone metadata
- The backend already has code to resolve annotations and extract agents from tokens, would be a redundant code in the processing service to resolve the annotation. besides, we already did a database call to check if the annotation exists, so performance isn't impacted, i just repurposed that call to provide more information

Other changes
- `getUser()` -> `getAgent()` from token, `User` object is obsolete

### DIFF
--- a/src/main/java/eu/dissco/backend/client/AnnotationClient.java
+++ b/src/main/java/eu/dissco/backend/client/AnnotationClient.java
@@ -3,6 +3,7 @@ package eu.dissco.backend.client;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import eu.dissco.backend.domain.annotation.batch.AnnotationEvent;
+import eu.dissco.backend.schema.Agent;
 import eu.dissco.backend.schema.Annotation;
 import org.springframework.cloud.openfeign.FeignClient;
 import org.springframework.web.bind.annotation.DeleteMapping;
@@ -25,6 +26,6 @@ public interface AnnotationClient {
       @PathVariable("suffix") String suffix, @RequestBody Annotation annotation);
 
   @DeleteMapping(value = "/{prefix}/{suffix}")
-  void deleteAnnotation(@PathVariable("prefix") String prefix,
-      @PathVariable("suffix") String suffix);
+  void tombstoneAnnotation(@PathVariable("prefix") String prefix,
+      @PathVariable("suffix") String suffix, Annotation annotation, Agent tombstoningAgent);
 }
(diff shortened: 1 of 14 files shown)


src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
```diff
@@ -39,13 +40,13 @@ public List<Annotation> getAnnotations(int pageNumber, int pageSize) {
         .limit(pageSizePlusOne).offset(offset).fetch(this::mapToAnnotation);
   }
 
-  public int getAnnotationForUser(String id, String userId) {
-    return context.select(ANNOTATION.ID)
+  public Optional<Annotation> getAnnotationForUser(String id, String userId) {
```
samleeflang, line 43: Maybe rename method to getActiveAnnotationForUser (unrelevant to this PR I know)

samleeflang (approved):
Just a check and a not really related remark, looks good otherwise :medal_military:

src/main/java/eu/dissco/backend/client/AnnotationClient.java
```diff
@@ -25,6 +26,6 @@ JsonNode updateAnnotation(@PathVariable("prefix") String prefix,
       @PathVariable("suffix") String suffix, @RequestBody Annotation annotation);
 
   @DeleteMapping(value = "/{prefix}/{suffix}")
-  void deleteAnnotation(@PathVariable("prefix") String prefix,
-      @PathVariable("suffix") String suffix);
+  void tombstoneAnnotation(@PathVariable("prefix") String prefix,
+      @PathVariable("suffix") String suffix, Annotation annotation, Agent tombstoningAgent);
```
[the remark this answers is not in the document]
southeo: you're right, had to wrap it.

src/main/java/eu/dissco/backend/controller/AnnotationController.java
```diff
@@ -183,7 +183,6 @@ public ResponseEntity<JsonApiWrapper> getAnnotationVersions(@PathVariable("prefi
     return ResponseEntity.ok(versions);
   }
 
-  @PreAuthorize("isAuthenticated()")
```
samleeflang, line 186: I think this should be kept in place?

samleeflang (commented):
Sorry, one last thing

src/main/java/eu/dissco/backend/controller/AnnotationController.java
samleeflang: Ok yes looks like it is there. Sorry, just looked strange. Approving the PR

PR #135, opened 2024-09-26
Add proxies to ids passed to mongodb

southeo:
Mongodb prov entities have proxy in their ids. This change adds the correct proxy to the id in the controller

Mongodb indexes have also been added on these terms


samleeflang (approved):
Sweet, thanks

PR #148, opened 2024-11-11
tombstoning can be done by admin

southeo:
Tombstoning can be done by users with role dissco-admin

The `isAdmin()` check can be applied to other operations as use cases arise

Sonar is whining about 92.6% test coverage, it's just a log message

### DIFF
--- a/src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
+++ b/src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
@@ -49,6 +49,14 @@ public class AnnotationRepository {
         .fetchOptional(this::mapToAnnotation);
   }
 
+  public Optional<Annotation> getActiveAnnotation(String id) {
+    return context.select(ANNOTATION.asterisk())
+        .from(ANNOTATION)
+        .where(ANNOTATION.ID.eq(id))
+        .and(ANNOTATION.TOMBSTONED.isNull())
+        .fetchOptional(this::mapToAnnotation);
+  }
+
   public List<Annotation> getForTarget(String id) {
     return context.select(ANNOTATION.DATA)
         .from(ANNOTATION)
--- a/src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
+++ b/src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
@@ -40,21 +40,15 @@ public class AnnotationRepository {
         .limit(pageSizePlusOne).offset(offset).fetch(this::mapToAnnotation);
   }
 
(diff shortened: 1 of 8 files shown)


src/main/java/eu/dissco/backend/repository/AnnotationRepository.java
```diff
@@ -49,6 +49,14 @@ public Optional<Annotation> getActiveAnnotationForUser(String id, String userId)
         .fetchOptional(this::mapToAnnotation);
   }
 
+  public Optional<Annotation> getActiveAnnotation(String id) {
```
samleeflang, line 52: I would probably have combined this with the one above and made the `.and(ANNOTATION.CREATOR.eq(userId))` in the condition

src/test/java/eu/dissco/backend/controller/AnnotationControllerTest.java
```diff
@@ -281,11 +288,40 @@ void testGetAnnotationsByVersion() throws NotFoundException {
     then(service).should().getAnnotationVersions(eq(HANDLE + ID), anyString());
   }
 
+  @ParameterizedTest
+  @MethodSource("nonAdminClaims")
+  void testTombstoneAnnotationSuccessNonAdmin(Map<String, Object> claims) throws Exception {
```
samleeflang, line 293: Is this the correct name, one test case seems like it is an Admin

samleeflang (commented):
Nice, just a small question/comment

src/main/java/eu/dissco/backend/service/AnnotationService.java
```diff
@@ -205,10 +205,10 @@ public boolean tombstoneAnnotation(String prefix, String suffix, Agent agent, bo
     Optional<Annotation> result;
     if (isAdmin) {
       log.info("Admin tombstoning annotation {}", id);
-      result = repository.getActiveAnnotation(id);
+      result = repository.getActiveAnnotation(id, null);
```
samleeflang, line 208: hmm not sure about passing null, why not do the isAdmin and then check on `!isAdmin` instead of the null check. Anyway this is not wrong

samleeflang (approved):
:man_cook:  Would have passed on the isAdmin myself but is works as well

src/main/java/eu/dissco/backend/service/AnnotationService.java
southeo: if `isAdmin=true` then the userId is going to be null anyway, thought this was a bit more concise. 
The alternative is to have two functions but someone suggested only having one :)

PR #155, opened 2024-11-29
use human readable type for responses

southeo:
we have 2 type fields:
- `@type`, which is human-readable
- `ods:fdoType`, which is a DOI

For the responses, we need to provide a type to adhere to jsonAPI standards.
```
{
   "data" : {
      "type" : "ods:DigitalSpecimen",   // This type here
      "attributes" : {
             ...
      }
   }
}

```

 We had some objects that returned fdo types, some that returned human-readable types. Now, we return the human readable for the type. this is because we have some response types (e.g. `digitalSpecimenVersions`) that don't have an FDO profile. 

[Jira - back end types should be consistent](https://naturalis.atlassian.net/browse/DD-1509?atlOrigin=eyJpIjoiNGNlMmJmY2MzNzRkNGY1NjkzYjZhMjIxNDIxMzhjYmIiLCJwIjoiaiJ9)


src/main/java/eu/dissco/backend/domain/FdoType.java
```diff
@@ -0,0 +1,39 @@
+package eu.dissco.backend.domain;
+
+import com.fasterxml.jackson.annotation.JsonAlias;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Getter;
+
+@Getter
+public enum FdoType {
```
samleeflang, line 8: Ok this list is final right? This is what we will use when we move to production

samleeflang (approved):
Good clean up, check if this impacts Tom in any way
:soap:

src/main/java/eu/dissco/backend/domain/FdoType.java
southeo: those 2 things aren't the same 😆 It's what we'll use for production though

src/main/java/eu/dissco/backend/domain/FdoType.java
```diff
@@ -0,0 +1,39 @@
...
+  ANNOTATION("https://doi.org/21.T11148/cf458ca9ee1d44a5608f", "ods:Annotation"),
+  @JsonProperty("https://doi.org/21.T11148/417a4f472f60f7974c12")
+  @JsonAlias("ods:SourceSystem")
+  SOURCE_SYSTEM("https://doi.org/21.T11148/417a4f472f60f7974c12", "ods:SourceSystem"),
+  @JsonProperty("https://doi.org/21.T11148/ce794a6f4df42eb7e77e")
+  @JsonAlias("ods:DataMapping")
+  DATA_MAPPING("https://doi.org/21.T11148/ce794a6f4df42eb7e77e", "ods:DataMapping"),
+  @JsonProperty("https://doi.org/21.T11148/22e71a0015cbcfba8ffa")
+  @JsonAlias("ods:MachineAnnotationService")
+  MAS("https://doi.org/21.T11148/22e71a0015cbcfba8ffa", "ods:MachineAnnotationService"),
```
samleeflang, line 26: Sorry it should be this:
https://dtr-test.pidconsortium.net/#objects/21.T11148/a369e128df5ef31044d4

PR #161, opened 2025-01-14
elvis services

southeo:
Adds ELViS endpoints

`/elvis/specimen/{prefix}/{suffix}` -> Given a DOI, retrieves exactly one specimen according to ELViS specifications
`/elvis/specimen/suggest/inventoryNumber` -> conducts wildcard search (i.e. a partial match) on provided searchValue. Checks input against `ods:physicalSpecimenID` and `dcterms:identifier` (doi)

[openAPI specification](https://github.com/user-attachments/files/18493413/api-docs.txt) shows schema and describes the mapping from openDS (see specifically the `elvis-controller`).

### DIFF
--- a/src/main/java/eu/dissco/backend/service/ElvisService.java
+++ b/src/main/java/eu/dissco/backend/service/ElvisService.java
@@ -1,0 +1,118 @@
+package eu.dissco.backend.service;
+
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import eu.dissco.backend.domain.elvis.ElvisSpecimen;
+import eu.dissco.backend.domain.elvis.InventoryNumberSuggestion;
+import eu.dissco.backend.properties.ApplicationProperties;
+import eu.dissco.backend.repository.DigitalSpecimenRepository;
+import eu.dissco.backend.repository.ElasticSearchRepository;
+import eu.dissco.backend.schema.DigitalSpecimen;
+import eu.dissco.backend.schema.TaxonIdentification;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import lombok.RequiredArgsConstructor;
+import org.apache.commons.lang3.tuple.Pair;
+import org.springframework.stereotype.Service;
+
+@Service
+@RequiredArgsConstructor
+public class ElvisService {
+
+  private final DigitalSpecimenRepository repository;
+  private final ElasticSearchRepository elasticSearchRepository;
+  private final ApplicationProperties applicationProperties;
+  private final ObjectMapper mapper;
+
+  public ElvisSpecimen searchByDoi(String id) {
+    var specimen = repository.getLatestSpecimenById(id);
+    return buildElvisSpecimen(specimen, applicationProperties.getBaseUrl());
+  }
+
+  public JsonNode searchBySpecimenId(String inventoryNumber, int pageNumber, int pageSize)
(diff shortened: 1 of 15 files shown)


src/main/java/eu/dissco/backend/service/ElvisService.java
```diff
@@ -0,0 +1,118 @@
...
+      DigitalSpecimen digitalSpecimen) {
+    if (digitalSpecimen.getOdsHasIdentifications().isEmpty()
+        || digitalSpecimen.getOdsHasIdentifications().get(0).getOdsHasTaxonIdentifications()
+        .isEmpty()) {
+      return Optional.empty();
+    }
+    // First, try to find accepted id
+    for (var identification : digitalSpecimen.getOdsHasIdentifications()) {
+      if (Boolean.TRUE.equals(identification.getOdsIsVerifiedIdentification())) {
+        if (identification.getOdsHasTaxonIdentifications().isEmpty()) {
```
southeo, line 105: Do i need to cover the following case: Accepted identification does not contain any taxonIdentifications, but a non-accepted Identification has a taxonIdentification. Seems like that shouldn't happen

southeo:
> Hi Soulaine,
> 
> Thanks for your work, it looks perfect to me.
> 
> Just a quick question, the suggest API accept both inventory number (as in Muse) AND DOI but the specimen API does only accept (from what I read) the DOI in input.
> 
> As we are not sure that we'll be able to add DOI for the may version, can you check if the specimen API accepts an inventory number (e.g. MNHN-P-P01899395) please?
> 
> Thanks again for your work :)
> 
> Have a nice day!
> 
> Maxime,

Hi Maxime, 

Thanks for looking it over!

The `suggest` endpoint takes one input, the "suggestValue". Then in the background we use elasticsearch to check both the DOI and the physicalSpecimenID (aka inventory number) for matches. 

I use a wildcard search for this, so we check partial matches 
- `dcterms:identifier = https://doi.org/*{searchValue}*` (I add the "https://doi.org/" in the code, so it should not be included in the request)
- `ods:physicalSpecimenId = *{searchValue}*`

So the value search value "99395" would return MNHN-P-P01899395 and other physical specimen ids with that substring. You can also use this endpoint to search for DOIs, but you're right, users are more likely to know the physical speicmen id.

southeo:
Sorry, I misread your question

> Hi Soulaine,
> 
> Thanks for your work, it looks perfect to me.
> 
> Just a quick question, the suggest API accept both inventory number (as in Muse) AND DOI but the specimen API does only accept (from what I read) the DOI in input.
> 
> As we are not sure that we'll be able to add DOI for the may version, can you check if the specimen API accepts an inventory number (e.g. MNHN-P-P01899395) please?
> 
> Thanks again for your work :)
> 
> Have a nice day!
> 
> Maxime,

Hi Maxime, 

Thanks for looking at this!

Our term "physicalSpecimenID" is the closest concept we have to the inventory, so I'll use that term. 

The specimen endpoint can't accept a physical specimen identifier as an argument, because the physical specimen identifer isn't unique within DiSSCo. If it did accept a physical speicmen ID as the argument, it would have to return a list of potential specimens, and that's the role of the `suggest` endpoint. 

However, the `suggest` endpoint returns the physical speicmen id and the DOI. In this endpoint, catalogNumber contains the physicalSpecimenID, and inventoryNumber is the DiSSCo DOI. (Not a perfect mapping, but it adheres to the existing structure). That way, the result of the suggest endpoint can still be used as input for the specimen endpoint. How does that sound to you? 

Soulaine

PR #164, opened 2025-03-18
Elastic Search Improvements - Wildcard support, documentation, adds "habitat" search term

southeo:
**Changes**
- In the swagger endpoint, documents which terms are searchable in the /search endpoint
- makes "habitat" searchable (request from Senckenberg)
- Adds wildcard support

**Issue**: Search term "physicalSpecimenId" doesn't follow our naming convention. I could change this to "physicalSpecimenID", but I suspect that would affect the front end. I could create a redundant term, and support both physId and physID for a while, but we'd need to remember to remove the "Id" term later. 

**Issue**: we should make more terms searchable. I'm sure there's a way to leverage our json schemas some way to get a full list of terms and json paths instead of hardcoding terms.

### DIFF
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -1,16 +1,4 @@
-# October 14 2024
-# Issue in Spring OpenFeign, Need to upgrade commons-io to version 2.14.0
-CVE-2024-47554
-
-# Nov 13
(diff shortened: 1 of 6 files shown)


samleeflang (approved):
Ok yes nice.
Let's discuss how we could do wider queries in a refinement

PR #167, opened 2025-04-23
Switch Kafka to RabbitMQ

samleeflang:
Switch Kafka to RabbitMQ for the scheduling of MAS


src/main/java/eu/dissco/backend/service/RabbitMqPublisherService.java
```diff
@@ -0,0 +1,27 @@
...
+public class RabbitMqPublisherService {
+
+  @Value("${rabbitmq.mas-exchange-name:mas-exchange}")
+  private String exchangeName = "mas-exchange";
+
+  private final ObjectMapper mapper;
+  private final RabbitTemplate rabbitTemplate;
+
+  public void sendObjectToQueue(String routingKey, Object object) throws JsonProcessingException {
+    log.info("Sending to exchange with routing key: {} and with object: {}", routingKey, object);
```
southeo, line 23: if you're sending the whole object, maybe make it a debug?

southeo (approved):
:stew:

src/main/java/eu/dissco/backend/service/RabbitMqPublisherService.java
samleeflang: :+1:

PR #169, opened 2025-05-28
fix duplicated pagination information ! deployment changes

southeo:
Before, we manually added the pageNumber and pageSize (and query params if applicable) to the Links object. However, these values are present in `httpRequest.getQueryString()`. If the user included pagination parameters, this information would then be duplicated. The parameters would always be duplicated. 

This PR returns a links object without the duplicated user information. 

Deployment changes: 
applicationProperties.baseURL needs to include /api/.


samleeflang (approved):
:page_with_curl:

PR #170, opened 2025-06-18
jooq update ! DB changes 

southeo:
Jooq code generation needed to resync with database 

Note: Minimum version of jooq is now 3.19.18 to comply with spring's version of postgres. Couldn't immediately upgrade to latest version (3.20.x) so we can make a ticket for this.


samleeflang (approved):
:lady_beetle:

PR #172, opened 2025-07-03
Return 404 for missing objects

southeo:
Annotation, Digital Media, and Digital Specimens now return a 404 instead of a 500 NPE
[Jira](https://naturalis.atlassian.net/browse/DD-1852?atlOrigin=eyJpIjoiMjZlY2FlZDdjMjU0NGZlMDhmNzY0OThiMzA2NWY2MzgiLCJwIjoiaiJ9)

### DIFF
--- a/src/main/java/eu/dissco/backend/controller/DigitalMediaController.java
+++ b/src/main/java/eu/dissco/backend/controller/DigitalMediaController.java
@@ -84,7 +84,7 @@ public class DigitalMediaController extends BaseController {
   public ResponseEntity<JsonApiWrapper> getDigitalMediaObjectById(
       @Parameter(description = PREFIX_OAS) @PathVariable("prefix") String prefix,
       @Parameter(description = SUFFIX_OAS) @PathVariable("suffix") String suffix,
-      HttpServletRequest request) {
+      HttpServletRequest request) throws NotFoundException {
     var id = prefix + '/' + suffix;
     log.info("Received get request for multiMedia with id: {}", id);
     var multiMedia = service.getDigitalMediaById(id, getPath(request));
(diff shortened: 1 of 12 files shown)


src/main/java/eu/dissco/backend/controller/DigitalMediaController.java
```diff
@@ -190,8 +190,7 @@ public ResponseEntity<JsonApiListResponseWrapper> getMasJobRecordForMedia(
       @Parameter(description = JOB_STATUS_OAS) @RequestParam(required = false) JobState state,
       @Parameter(description = PAGE_NUM_OAS) @RequestParam(defaultValue = DEFAULT_PAGE_NUM) int pageNumber,
       @Parameter(description = PAGE_SIZE_OAS) @RequestParam(defaultValue = DEFAULT_PAGE_SIZE) int pageSize,
-      HttpServletRequest request
-  ) throws NotFoundException {
```
samleeflang, line 194: No NotFoundExcepion here? What if you call it for a media that does not exists?

src/main/java/eu/dissco/backend/controller/DigitalSpecimenController.java
samleeflang, line 1: Shouldn't the full endpoint also thrown a NotFound?

samleeflang (commented):
I think there are more endpoints that should throw a NotFoundException. Could you check, for example the specimen full or the media /mas etc...

src/main/java/eu/dissco/backend/controller/DigitalMediaController.java
southeo: the underlying db query is: 

```
SELECT * from mas_job_records
WHERE target_id == {target_id}
```

If this response is empty (it returns a list), it could either mean the media doesn't exist, or more often, there aren't any scheduled mas jobs for that media. We could add a check to see if the media exists, but I don't think that adding an extra db call is a good idea. 

Besides, because this returns a list, if the query is empty, we don't throw a NPE, which is what this PR is trying to avoid

samleeflang: Ok clear

PR #178, opened 2025-08-22
Use Search_After in ElasticSearch pagination

southeo:
# Changes to ElasticSearch

Because ElasticSearch doesn't support paginated queries over 10k objects, we need to change our
approach. We use the `search_after` API to cycle through a high volume of results.

This is only for searching specimens/annotations; aggregations stay the same.

1. Sort results by some field (I chose `dcterms:created`, and use `dcterms:identifier` as tiebreaker)
2. Check if we've reached the requested page
   a. If yes: stop searching, return result
   b. If no: take last value of the sorted fields (created and identifier)
3. Continue to search starting after the fields we took from step 2b

## Why also change annotations?

Currently, we only use ES to search by user (creator). I thought it would be good to use the
search_after for these queries, because we may want to extend this to searching by MAS creators,
which will likely far exceed the 10k limit.

## No `search_after` for ELViS?

There's an exception we'd need to consider for the ELViS endpoint. (It uses a should_ query instead
of a must_). Because we're removing the endpoint soon, I left the ELViS endpoint unchanged.

# Get Latest Specimens vs Get Any Specimens

We have 2 endpoints:

- GET specimens/v1/latest : returns latest specimens
- GET specimens/v1 : returns any specimens

Before, these two endpoints had different logic. The /latest endpoint would sort by date created,
and the regular endpoint wouldn't. Because we use the date created in our sort_by query, these two
endpoints do exactly the same thing now (and rely on the same method). The endpoints are distinct
for now, but we should remove the /latest endpoint at some point. I've marked it as deprecated for now. 

# Changes to Plus-one-to-change-next

Before, we increased the pageSize to check next in elastic queries. However, this actually results
in missed results.

If we have documents 1-18, and we query page Number = 3, page size = 5:

Expected pagination: [1 2 3 4 5] [6 7 8 9 10] [11 12 13 14 15] [16 17 18]

Expected results: [11 12 13 14 15]

If we increase page size by 1, however, pagination
becomes: [1 2 3 4 5 6] [7 8 9 10 11 12] [13 14 15 16 17 18]

Results we get: [13 14 15 16 17 18] (truncated to [13 14 15 16 17] for the user)

So we really don't want to increase the page size. (We probably also want to check this for DB
queries, but that's beyond the scope of this PR)

We can use the ElasticSearch `TotalHits` instead of increasing pageSize to check if there are more
results.

### DIFF
--- a/src/main/java/eu/dissco/backend/controller/DigitalSpecimenController.java
+++ b/src/main/java/eu/dissco/backend/controller/DigitalSpecimenController.java
@@ -78,12 +78,16 @@ public class DigitalSpecimenController extends BaseController {
     return ResponseEntity.ok(specimen);
   }
 
+  /**
+   * @deprecated
+   */
   @Operation(summary = "Get latest paginated digital specimens")
   @ApiResponses(value = {
       @ApiResponse(responseCode = "200", description = "Digital specimens successfully retrieved", content = {
           @Content(mediaType = "application/json", schema = @Schema(implementation = DigitalSpecimenResponseList.class))
       })
   })
+  @Deprecated(since = "1.1.3", forRemoval = true)
   @GetMapping(value = "/latest", produces = MediaType.APPLICATION_JSON_VALUE)
   public ResponseEntity<JsonApiListResponseWrapper> getLatestSpecimen(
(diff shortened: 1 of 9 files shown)


southeo:
Slow performance. More refinement is needed. We may need to get the "search_after" value from users instead.

PR #179, opened 2025-08-26
Some minor fixes which came out of functional testing

samleeflang:
- Have better roleName
- Fix the /versions to look at correct mongo field
- Add realm roles to security stuff

### DIFF
--- a/src/test/java/eu/dissco/backend/security/JwtAuthConverterTest.java
+++ b/src/test/java/eu/dissco/backend/security/JwtAuthConverterTest.java
@@ -69,7 +69,7 @@ class JwtAuthConverterTest {
     // Then
     assertThat(token.isAuthenticated()).isTrue();
     assertThat(token.getName()).isEqualTo("adf294ba-bb03-4962-8042-a37f1648458e");
-    assertThat(token.getAuthorities()).hasSize(0);
+    assertThat(token.getAuthorities()).isEmpty();
   }
 
 }


.github/workflows/.trivyignore
```diff
@@ -1,7 +1,3 @@
 # Jul 16 2025
 # Issue with commons-fileupload, needs an update of spring-cloud-openfeign
 CVE-2025-48976
```
southeo, line 3: Is this still relevant?

src/main/java/eu/dissco/backend/repository/MongoRepository.java
```diff
@@ -38,16 +39,16 @@ public JsonNode getByVersion(String id, int version, String collectionName)
         .get(PROV_VALUE);
   }
 
-  public List<Integer> getVersions(String id, String collectionName) throws NotFoundException {
+  public List<Integer> getVersions(String id, String collectionName, String versionProperty) throws NotFoundException {
```
southeo, line 42: Since `versionProperty` depends on the collection, it might be a bit safer to make a map to get the version property instead of having it be passed as a method param? Just a thought

southeo (approved):
:taco:

.github/workflows/.trivyignore
samleeflang: Yes checked and this is still an relevant (they are working on a new major version)

src/main/java/eu/dissco/backend/repository/MongoRepository.java
samleeflang: Changed it to an Enum, I think this looks better now.

PR #201, opened 2026-04-07
gitleaks config

southeo:
Adds gitleaks config to ignore specific false positive leak

Alternative would be to add /test dir path but I'd like to keep testing that directory just to be safe

also small bug fix to add a path variable to a http service client


samleeflang (approved):
🪲

PR #202, opened 2026-04-13
rest client fix

southeo:
https://naturalis.atlassian.net/browse/DD-3063


.github/workflows/build.yaml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
   schedule:
-    - cron: "0 4  * * MON" # run on Monday at 4 AM
+    - cron: "0 3 1 * *" # At 3:00 on the 1st day of the month
```
samleeflang, line 9: ok ok, if the CISRT member can do this I will also do this with any repo I touch

src/main/java/eu/dissco/backend/service/MasJobRecordService.java
```diff
@@ -85,25 +79,6 @@ private JsonApiListResponseWrapper packageList(List<MasJobRecordFull> masJobReco
 		return new JsonApiListResponseWrapper(dataList, linksNode);
 	}
 
-	public Map<String, MasJobRecord> createMasJobRecord(Set<MachineAnnotationService> masRecords, String targetId,
```
samleeflang, line 88: What is going on here? This isn't mentioned in the overview or ticket.

samleeflang (approved):
🦕

src/main/java/eu/dissco/backend/service/MasJobRecordService.java
southeo: trust me bro

PR #205, opened 2026-05-18
Media derivative behind feature toggle

southeo:
Moves media derivatives behind a feature toggle, as this feature is not ready for production yet. 

I don't want to remove the s3 properties validation, but those buckets don't exist for the production environment yet. Instead, I create the dependent beans conditionally, based on the profile. So if profile does not include MEDIA_DERIVATIVE, then we don't create S3Properties, S3Configuration, or S3Repository Beans. 

There are now 2 separate configurations for the DigitalMediaService, depending on active profile.


samleeflang (dismissed):
🖼️

PR #207, opened 2026-05-26
Feature toggles for Media Derivatives ! Deployment change

southeo:
**Remove digital media config**

The config we had wasn't working for the `default` profile. That's because annotating DigitalMediaService with the @Service annotation already created a bean, which wasn't allowing the 

Instead of having 2 constructors, I marked the s3 as nullable. It will only be null in the `media-derviative` profile. 

**Prevent users from using /thumbnail endpoint**
Removed /thumbnail support outside media-derivative profile. This is related to the media derivative functionality and is part of the profile

**Harmonized feature toggles** 
We determined if the backend supports accepting annotations by a property in the `ApplicationProperties` class (i.e. `application.isAcceptingAnnotations` in application.properties)

We want all features to be toggled under the same mechanism. I moved the above property to a profile instead. 

Spring will parse `spring.profiles.active=default,accept-annotations` as a list. This just needs a small deployment change to harmonize.


samleeflang (approved):
🧷 Sounds good

